PR #45, opened 2026-02-07
Add install/uninstall commands with embedded manifests

### LINKED ISSUE #29: axon install/uninstall feature
Add install/uninstall commands to axon to simplify the installation

gjkim42:
## Summary
- Add `axon install` and `axon uninstall` commands using Kubernetes client-go SDK
- Embed installation YAML manifests into the binary using `go:embed` (works with `go install`)
- Use dynamic client with server-side apply (ApplyPatchType) for creating resources
- Use dynamic client with standard delete for removing resources, with NotFound/NoMatch tolerance for idempotent uninstalls
- Parse multi-document YAML into unstructured objects using discovery-based REST mapper to resolve GVRs
- REST mapper is created once per manifest set (outside the loop) to avoid redundant API server calls
- Update CI workflow to use `axon install` instead of `kubectl apply`
- Update README.md with install/uninstall documentation
- Add unit tests covering YAML parsing, embedded manifests, config skip, and argument rejection
- Add integration tests using envtest for install/uninstall lifecycle

Closes #29

## Test plan
- [x] `make test` passes (unit tests for YAML parsing, embedded manifests, config skip, argument rejection)
- [x] `make test-integration` passes (install/uninstall integration tests with envtest)
- [ ] `axon install` applies embedded CRDs then controller to a real cluster
- [ ] `axon uninstall` removes controller then CRDs from a real cluster
- [ ] `axon install --kubeconfig /path/to/config` uses custom kubeconfig
- [ ] `axon install extra-arg` returns an error
- [ ] `axon uninstall` is idempotent (no errors on already-uninstalled cluster)
- [ ] Works with `go install github.com/gjkim42/axon/cmd/axon@latest`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -82,16 +82,15 @@ jobs:
           kind load docker-image gjkim42/axon-spawner:e2e
           kind load docker-image gjkim42/claude-code:e2e
 
+      - name: Build CLI
+        run: make build WHAT=cmd/axon
+
       - name: Deploy controller
         run: |
-          kubectl apply -f install-crd.yaml
-          kubectl apply -f install.yaml
+          bin/axon install
           kubectl set image deployment/axon-controller-manager manager=gjkim42/axon-controller:e2e -n axon-system
           kubectl patch deployment axon-controller-manager -n axon-system -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager","imagePullPolicy":"Never","args":["--leader-elect","--spawner-image=gjkim42/axon-spawner:e2e","--spawner-image-pull-policy=Never","--claude-code-image=gjkim42/claude-code:e2e","--claude-code-image-pull-policy=Never"]}]}}}}'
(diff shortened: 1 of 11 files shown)


gjkim42:
you cannot guarantee that user uses axon install in our git repository.
I prefer to archive our installation yaml to the binary.

gjkim42:
I think we cannot guarantee that users have kubectl installed. Why don't we just use kubernetes sdk directly?

gjkim42:
Yes, `go install github.com/gjkim42/axon/cmd/axon@latest` works correctly. The `go:embed` directive embeds the YAML manifests into the binary at compile time. Since `install-crd.yaml` and `install.yaml` are checked into `internal/manifests/`, they are available to the Go toolchain during `go install` and get compiled into the binary.

Also fixed the failing tests — the fake kubectl scripts use `cat` to consume stdin, but setting `PATH` to only the temp directory made `cat` unavailable in CI. The fix prepends the temp directory to PATH (so our fake kubectl takes priority) instead of replacing PATH entirely, keeping system utilities accessible. The `RequiresKubectl` tests still fully override PATH since they intentionally need kubectl to be missing.

PR #69, opened 2026-02-07
Add label workflow to apply needs-triage on new issues

### LINKED ISSUE #67: add needs-triage if there's no triage-accepted label
you can use github workflow to do this

gjkim42:
## Summary
- Adds a GitHub Actions workflow (`label.yaml`) that automatically manages the `needs-triage` label on issues
- Checks current labels on each event rather than relying on event types for reliability
- Adds `needs-triage` to newly opened issues that don't have `triage-accepted`
- Removes `needs-triage` when `triage-accepted` is added to an issue
- Re-adds `needs-triage` when `triage-accepted` is removed from an issue

Fixes #67

## Test plan
- [ ] Open a new issue without `triage-accepted` - verify `needs-triage` is added
- [ ] Open a new issue with `triage-accepted` already set - verify `needs-triage` is NOT added
- [ ] Add `triage-accepted` to an issue with `needs-triage` - verify `needs-triage` is removed
- [ ] Remove `triage-accepted` from an issue - verify `needs-triage` is re-added

### DIFF
--- a/.github/workflows/label.yaml
+++ b/.github/workflows/label.yaml
@@ -1,0 +1,33 @@
+name: Label
+
+on:
+  issues:
+    types: [opened, labeled, unlabeled]
+
+jobs:
+  triage-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add or remove needs-triage label
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const labels = context.payload.issue.labels.map(l => l.name);
+            const hasTriageAccepted = labels.includes('triage-accepted');
+            const hasNeedsTriage = labels.includes('needs-triage');
+
+            if (!hasTriageAccepted && !hasNeedsTriage) {
+              await github.rest.issues.addLabels({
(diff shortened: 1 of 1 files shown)


.github/workflows/triage.yaml
```diff
@@ -0,0 +1,23 @@
+name: Triage
```
gjkim42: This will not be just for triage. Name it Label, or any recommendation?

gjkim42: 🤖 **Axon Agent** @gjkim42

Renamed the workflow from `triage.yaml` to `label.yaml` and updated the workflow name from "Triage" to "Label". This makes it more general-purpose for future label management rules.

gjkim42:
you have to remove needs-triage label if anyone labels triage-accepted label and also you have to re-add needs-triage label if anyone removes triage-accepted label

gjkim42:
Checking the current labels regardless of the event would be more reliable way like #79

PR #132, opened 2026-02-08
Fix workspace ref lookup race with informer cache

### LINKED ISSUE #123: Fix flaky integration test: workspace ref lookup races with informer cache
## Problem

The integration test "Should create a Job with init container and workspace volume" (`test/integration/task_test.go:245`) is flaky. The `TaskReconciler` uses the cached client (`r.Get`) to look up the Workspace resource. When the Workspace is just created, the informer cache may not have synced yet, causing a NotFound error. The controller then immediately marks the Task as `Failed` and stops retrying.

## Relevant code

`internal/controller/task_controller.go:144-158` — when `r.Get` returns NotFound for the Workspace, the Task is permanently failed with no requeue.

## Considerations

- Simply requeuing on NotFound would fix the race but means truly missing workspaces are never reported as failed
- Using `APIReader` (direct API server read) avoids the cache race but adds direct API load
- Could add a retry count / backoff before marking as permanently failed

gjkim42:
## Summary

- Requeue on Workspace `NotFound` instead of marking the Task as `Failed`, fixing the informer cache race
- When the cached `r.Get()` returns NotFound for a Workspace, requeue after 2s to let the cache sync
- This also handles the case where a Workspace is created after the Task: the controller retries until the Workspace appears

## Changes

- `internal/controller/task_controller.go`: Replace immediate `Failed` status on Workspace NotFound with `RequeueAfter: 2s`
- `test/integration/task_test.go`: Update integration test for nonexistent workspace to verify no Job is created while the workspace is missing, the Task is not marked as Failed, and the Job is created once the Workspace appears

Fixes #123

## Test plan

- [x] `make test` — unit tests pass
- [x] `make test-integration` — all 24 integration tests pass (including updated nonexistent workspace test)
- [ ] CI checks pass

### DIFF
--- a/internal/controller/task_controller.go
+++ b/internal/controller/task_controller.go
@@ -147,16 +147,11 @@ func (r *TaskReconciler) createJob(ctx context.Context, task *axonv1alpha1.Task)
 			Namespace: task.Namespace,
 			Name:      task.Spec.WorkspaceRef.Name,
 		}, &ws); err != nil {
-			logger.Error(err, "Unable to fetch Workspace", "workspace", task.Spec.WorkspaceRef.Name)
 			if apierrors.IsNotFound(err) {
-				task.Status.Phase = axonv1alpha1.TaskPhaseFailed
-				task.Status.Message = fmt.Sprintf("Workspace %q not found", task.Spec.WorkspaceRef.Name)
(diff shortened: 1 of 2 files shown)


gjkim42:
P0: I really don't want to directly read resources from the API server.
is there a better way to do that?

PR #166, opened 2026-02-08
Support versioned releases with tagged container images

### LINKED ISSUE #165: stable release
Currently, we only install the latest version of axon.

I'd like to have a stable release.
- make image for each release
- update axon install to install the fixed version
- do we need more?? TBD

gjkim42:
## Summary
- Add `internal/version` package with build-time version injection via ldflags
- Add `axon version` subcommand to print the current binary version
- Update `axon install` to rewrite `:latest` image tags in manifests to match the CLI version (e.g. `:v0.1.0`), so versioned binaries deploy matching images
- Update Makefile `build` target to pass `-ldflags` with the `VERSION` variable
- Update release workflow to trigger on `v*` tags and push both versioned and `:latest` images

### Releasing a new version
1. Tag a commit: `git tag v0.1.0 && git push origin v0.1.0`
2. The release workflow builds and pushes images tagged `v0.1.0` and `latest`
3. Users install with: `go install github.com/axon-core/axon/cmd/axon@v0.1.0`
4. Running `axon install` deploys the controller with images tagged `v0.1.0`

### Development builds
- `make build` (no VERSION) produces a binary with version `latest`
- `axon install` uses the embedded manifests as-is (`:latest` tags)

## Test plan
- [x] Unit tests for `versionedManifest` with latest, tagged, and multi-image cases
- [x] Unit test for `axon version` command
- [x] Verified `make build VERSION=v0.1.0 WHAT=cmd/axon && bin/axon version` outputs `v0.1.0`
- [x] `make test` passes
- [x] `go vet ./...` passes

Fixes #165

---
## Summary by cubic
Adds build-time and runtime version detection so axon install deploys images that match the CLI (including go install builds). Updates the release workflow to publish versioned and latest images. Fixes #165.

- **New Features**
  - Version detection via ldflags or Go build info; only clean semver accepted; adds `axon version`.
  - `axon install` rewrites `:latest` to `:vX.Y.Z`; keeps `latest` for dev builds.
  - Release workflow builds/pushes versioned and `latest` tags on `v*`; Makefile injects ldflags only when VERSION != latest.

- **Dependencies**
  - Promote golang.org/x/mod to a direct dependency for semver checks.

<sup>Written for commit 64a41a0cefe97d32ce7ed9148005fc56bd3c6ba8. Summary will update on new commits.</sup>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/onsi/gomega v1.38.3
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/spf13/cobra v1.10.2
+	golang.org/x/mod v0.31.0
 	k8s.io/api v0.35.0
 	k8s.io/apimachinery v0.35.0
 	k8s.io/client-go v0.35.0
@@ -60,7 +61,6 @@ require (
 	go.uber.org/zap v1.27.0 // indirect
 	go.yaml.in/yaml/v2 v2.4.3 // indirect
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
-	golang.org/x/mod v0.31.0 // indirect
 	golang.org/x/net v0.48.0 // indirect
 	golang.org/x/oauth2 v0.30.0 // indirect
 	golang.org/x/sync v0.19.0 // indirect


Makefile
```diff
@@ -75,7 +79,7 @@ verify: controller-gen ## Verify everything is up-to-date and correct.
 build: ## Build binaries (use WHAT=cmd/axon to build specific binary).
 	@for dir in $$(go list ./$(or $(WHAT),cmd/...)); do \
 		bin_name=$$(basename $$dir); \
-		CGO_ENABLED=0 go build -o bin/$$bin_name $$dir; \
+		CGO_ENABLED=0 go build -ldflags "$(LDFLAGS)" -o bin/$$bin_name $$dir; \
```
gjkim42, line 85: does it work if we install axon cli with go install??
I guess not... do we have a better idea?

gjkim42:
🤖 **Axon Agent** @gjkim42

Addressed the review feedback about `go install` not picking up ldflags.

**Changes made:**
- Added an `init()` function in `internal/version` that falls back to `runtime/debug.ReadBuildInfo()` when ldflags aren't applied. When a user installs via `go install github.com/axon-core/axon/cmd/axon@v0.1.0`, Go embeds the module version in the binary, and `ReadBuildInfo()` returns `"v0.1.0"`.
- Uses `golang.org/x/mod/semver` to only accept clean semver tags, filtering out pseudo-versions from local builds (e.g. `v0.0.0-20260101-abcdef`).
- Updated the Makefile to skip ldflags when `VERSION=latest` (the default), so dev builds correctly fall through to the `init()` logic.

**Version resolution priority:**
1. ldflags (from `make build VERSION=v0.1.0`) — highest priority
2. `runtime/debug.ReadBuildInfo()` (from `go install ...@v0.1.0`) — fallback
3. `"latest"` — final default for dev builds

All tests pass.

PR #187, opened 2026-02-09
Wait for terminating resources before server-side apply

### LINKED ISSUE #185: local-run.sh may remove CRDs
I ran local-run.sh to re-install axon, but the CRDs are removed...

Find out the root cause and fix it.

gjkim42:
## Summary
- Fix `axon install` to wait for resources that have a `deletionTimestamp` before applying via server-side apply
- When a CRD is being deleted (has a `deletionTimestamp`), server-side apply succeeds but the CRD continues to be garbage collected, leaving the API unavailable
- This caused CRDs to be removed when re-running `local-run.sh` shortly after an uninstall

## Changes
- Added `waitForDeletion()` helper that polls until a resource is fully deleted
- Updated `applyManifests()` to check each resource for a `deletionTimestamp` before applying, waiting for deletion to complete if found
- Simplified `restoreCRDs()` in integration tests since the install command now handles the CRD deletion race internally
- Added unit tests for `waitForDeletion`

## Test plan
- [x] Unit tests pass (`make test`)
- [x] Integration tests pass (`make test-integration`) - confirmed the fix is active (logs show "Waiting for CustomResourceDefinition taskspawners.axon.io to be deleted")
- [x] Verification checks pass (`make verify`)

Fixes #185

---
## Summary by cubic
Waits for terminating resources before server-side apply in axon install, preventing CRDs from being re-applied while still deleting and avoiding API downtime. Fixes Linear #185.

- **Bug Fixes**
  - Added waitForDeletion (500ms poll, 60s timeout) and updated applyManifests to GET and pause when a resource has a deletionTimestamp.
  - Logs “Waiting for … to be deleted” and only applies after deletion completes, avoiding CRD GC after reinstall.
  - Simplified integration CRD restore; added unit tests for already-gone and eventually-deleted cases.

<sup>Written for commit 4382ca9b9ab290388e2f96ea178dfa091436e310. Summary will update on new commits.</sup>

### DIFF
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"time"
 
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -14,6 +15,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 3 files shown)


gjkim42:
So... I had tried to delete the CRD, and they actually were deleted on axon install?
Is this what you are saying?

PR #236, opened 2026-02-10
Add CRD-level enum validation for agent type field

### LINKED ISSUE #230: CLI: Missing validation for --type and --credential-type flags allows invalid values
## Problem

The CLI accepts invalid values for `--type` and `--credential-type` flags without any client-side validation. The error only occurs when Kubernetes rejects the resource.

### Reproduction

```bash
# Invalid agent type - no client-side error
./bin/axon run -p "test" --secret test --type invalid-type --dry-run
# Output: Creates YAML with type: invalid-type (which would fail in k8s)

# Invalid credential type - no client-side error
./bin/axon run -p "test" --secret test --credential-type wrong --dry-run
# Output: Creates YAML with type: wrong (which would fail in k8s)
```

Both commands succeed and generate YAML, but the values would be rejected by the CRD validation in Kubernetes.

## Expected Behavior

The CLI should validate enum values client-side and provide helpful error messages:

```bash
$ axon run -p "test" --secret test --type invalid-type
Error: invalid agent type "invalid-type". Valid types: claude-code, codex, gemini

$ axon run -p "test" --secret test --credential-type wrong  
Error: invalid credential type "wrong". Valid types: api-key, oauth
```

## Impact

**Poor user experience:**
- Users discover errors late (after kubectl apply, not during `axon run`)
- Error messages from Kubernetes are less helpful than CLI errors
- Users waste time debugging Kubernetes errors instead of fixing typos
- The `--dry-run` flag doesn't catch these issues

**Inconsistent with other validation:**
- The CLI validates required fields like `--prompt`
- But doesn't validate enum fields

## Implementation Notes

The valid values are defined in `api/v1alpha1/task_types.go`:
- `CredentialType`: `api-key`, `oauth` (line 10-15)
- Agent types are not explicitly enumerated in the CRD but common values are: `claude-code`, `codex`, `gemini`

The CLI should:
1. Define these enums in the CLI package
2. Validate flag values before creating resources
3. Show valid options in error messages
4. Optionally: List valid values in `--help` text (e.g., `--type string    agent type (claude-code, codex, gemini) (default "claude-code")`)

## Related Issues

- #228: Similar issue about client-side validation
- #182: CLI validation gaps

gjkim42:
## Summary
- Add CRD-level `+kubebuilder:validation:Enum=claude-code;codex;gemini` markers on `TaskSpec.Type` and `TaskTemplate.Type` so Kubernetes rejects invalid agent types at admission time
- Update flag help text to list valid values inline (e.g., `agent type (claude-code, codex, gemini)`)
- Removed client-side validation per review feedback; server-side CRD validation is sufficient

## Test plan
- [x] `make verify` passes
- [x] `make test` passes
- [x] CRD YAML correctly includes enum constraints for agent type fields

Fixes #230

### DIFF
--- a/internal/cli/completion_test.go
+++ b/internal/cli/completion_test.go
@@ -104,6 +104,25 @@ func TestFlagCompletionOutput(t *testing.T) {
 	}
 }
 
+func TestFlagCompletionAgentType(t *testing.T) {
+	root := NewRootCommand()
+
+	root.SetArgs([]string{"__complete", "run", "--type", ""})
+	out := &strings.Builder{}
+	root.SetOut(out)
+	root.Execute()
+
+	output := out.String()
+	for _, expected := range []string{"claude-code", "codex", "gemini"} {
+		if !strings.Contains(output, expected) {
+			t.Errorf("expected %s in type completions, got %q", expected, output)
+		}
+	}
+	if !strings.Contains(output, ":4") {
+		t.Errorf("expected ShellCompDirectiveNoFileComp (:4) in output, got %q", output)
+	}
+}
+
 func TestFlagCompletionCredentialType(t *testing.T) {
 	root := NewRootCommand()
 
--- a/internal/cli/dryrun_test.go
+++ b/internal/cli/dryrun_test.go
@@ -65,6 +65,49 @@ func TestRunCommand_DryRun(t *testing.T) {
 	}
 }
 
+func TestRunCommand_DryRun_AgentType(t *testing.T) {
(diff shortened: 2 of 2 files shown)


internal/cli/create.go
```diff
@@ -164,6 +164,10 @@ func newCreateTaskSpawnerCommand(cfg *ClientConfig) *cobra.Command {
 				}
 			}
 
+			if err := validateCredentialType(credentialType); err != nil {
```
gjkim42: We validate it using crd validation.

Do we need to add this? It will end up getting errors from the ser er.

PR #785, opened 2026-03-24
feat: native IRSA auth

### LINKED ISSUE #780: Allow for authentication via bedrock instead of Anthropic directly
We would like to use Kelos but would like it to call models directly via bedrock rather than via the anthropic API. Seeing as it is running the claude-sdk which should support this out of the box, we believe that allowing this as a authentication model is worthwhile

knechtionscoding:
#### What type of PR is this?

/kind feature

Add a generic `none` credential type that disables built-in credential injection, allowing users to supply their own provider credentials via `podOverrides.env` and `podOverrides.serviceAccountName`. This enables AWS Bedrock, Google Vertex AI, Azure OpenAI, or any other provider without coupling the CRD to provider-specific concepts.

Also refactors credential injection into a centralized `credentialEnvVars()` function and moves `ServiceAccountName` to `PodOverrides` where it naturally belongs (useful for any workload identity pattern, not just credentials).

Changes:
- Add `CredentialTypeNone` constant and update CRD enum (`api-key`, `oauth`, `none`)
- Make `Credentials.SecretRef` a pointer (`*SecretReference`) — optional for `none`, required for `api-key`/`oauth` (enforced via CEL validation)
- Add `PodOverrides.ServiceAccountName` for workload identity (IRSA, GKE WI, Azure WI)
- Add `credentialEnvVars()` to centralize credential env var injection for all types
- Remove provider-specific fields (`Region`, `ServiceAccountName`) from `Credentials`
- Remove `BedrockConfig` from CLI config and `--region`/`--service-account` flags
- Add `--credential-type none` support in CLI
- Add unit tests and dry-run tests for `none` credential type
- Add example manifests in `examples/09-bedrock-credentials/` showing Bedrock via `none` + PodOverrides

Fixes #780

- **API breaking change**: `Credentials.SecretRef` changed from value type to pointer. All existing manifests with `secretRef` set continue to work — the CRD is backwards-compatible. CEL validation ensures `secretRef` is still required for `api-key` and `oauth` types.
- The `credentialEnvVars()` refactor is a no-op for existing `api-key` and `oauth` types — it just moves the same logic into a single function.
- `ServiceAccountName` lives on `PodOverrides` rather than `Credentials` because it controls pod identity for any reason (GCS, S3, cloud APIs), not just credentials.
- The `none` type is provider-agnostic by design — users bring their own env vars via `podOverrides.env` and `podOverrides.serviceAccountName`. No provider-specific fields leak into the CRD.

```release-note
Add `none` credential type (`credentials.type: none`) that disables built-in credential injection. Users supply provider-specific credentials via `podOverrides.env` and can set `podOverrides.serviceAccountName` for workload identity (IRSA, GKE Workload Identity, Azure Workload Identity). This enables AWS Bedrock, Google Vertex AI, and other providers without provider-specific CRD fields.
```


api/v1alpha1/task_types.go
```diff
@@ -13,6 +13,8 @@ const (
 	CredentialTypeAPIKey CredentialType = "api-key"
 	// CredentialTypeOAuth uses OAuth for authentication.
 	CredentialTypeOAuth CredentialType = "oauth"
+	// CredentialTypeBedrock uses AWS credentials for Bedrock authentication.
+	CredentialTypeBedrock CredentialType = "bedrock"
```
gjkim42: The `bedrock` credential type feels too specific to Claude Code. If we later need Vertex, Azure OpenAI, or other providers, each would need its own type, CRD enum value, and code path in the controller.

We already have `PodOverrides.Env` which supports `valueFrom.secretKeyRef`, so users can inject any provider's env vars today. The only missing piece is `ServiceAccountName` on `PodOverrides` (for IRSA and similar workload identity patterns).

What do you think about replacing `bedrock` with a generic `none` credential type instead? It would mean:
- `type: none` — no credential injection by the controller; users bring their own via `PodOverrides`
- Move `ServiceAccountName` from `Credentials` to `PodOverrides` (useful for any workload identity, not just bedrock)
- CEL validation: `self.type == 'none' || has(self.secretRef)`

Example Task with bedrock via `none` + `PodOverrides`:
```yaml
credentials:
  type: none
podOverrides:
  serviceAccountName: bedrock-agent-sa
  env:
    - name: CLAUDE_CODE_USE_BEDROCK
      value: "1"
    - name: AWS_REGION
      value: us-east-1
```

Or with static credentials from a Secret:
```yaml
credentials:
  type: none
podOverrides:
  env:
    - name: CLAUDE_CODE_USE_BEDROCK
      value: "1"
    - name: AWS_ACCESS_KEY_ID
      valueFrom:
        secretKeyRef: { name: bedrock-creds, key: AWS_ACCESS_KEY_ID }
    - name: AWS_SECRET_ACCESS_KEY
      valueFrom:
        secretKeyRef: { name: bedrock-creds, key: AWS_SECRET_ACCESS_KEY }
    - name: AWS_REGION
      valueFrom:
        secretKeyRef: { name: bedrock-creds, key: AWS_REGION }
```

This makes the feature provider-agnostic from day one and avoids coupling the CRD to AWS-specific concepts like IRSA and region.

api/v1alpha1/task_types.go
```diff
@@ -39,12 +41,26 @@ type SecretReference struct {
...
+	// Region specifies the cloud provider region (e.g. AWS region for Bedrock).
+	// Used with bedrock credentials when secretRef is omitted (IRSA mode).
+	// +optional
+	Region string `json:"region,omitempty"`
+
+	// ServiceAccountName overrides the pod's service account.
+	// Use with IAM Roles for Service Accounts (IRSA) on EKS to let
+	// the pod assume an IAM role without static credentials.
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
```
gjkim42: `ServiceAccountName` isn't really credential-specific — it controls pod identity, which could be useful for non-credential reasons too (e.g., accessing GCS, S3, or other cloud services). `PodOverrides` seems like a more natural home for it, alongside `NodeSelector` and `Resources`.

api/v1alpha1/task_types.go
```diff
@@ -39,12 +41,26 @@ type SecretReference struct {
...
-	SecretRef SecretReference `json:"secretRef"`
+	// Required for api-key and oauth types. Optional for bedrock
+	// when using IAM Roles for Service Accounts (IRSA).
+	// +optional
+	SecretRef *SecretReference `json:"secretRef,omitempty"`
+
+	// Region specifies the cloud provider region (e.g. AWS region for Bedrock).
+	// Used with bedrock credentials when secretRef is omitted (IRSA mode).
+	// +optional
+	Region string `json:"region,omitempty"`
```
gjkim42: Same concern — `Region` is an AWS/provider concept leaking into the generic `Credentials` struct. With a `none` type + `PodOverrides.Env`, users would just set `AWS_REGION` as an env var, which is more explicit and doesn't require the CRD to model provider-specific fields.

api/v1alpha1/task_types.go
```diff
@@ -84,6 +100,7 @@ type TaskSpec struct {
 
 	// Credentials specifies how to authenticate with the agent.
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self.type == 'bedrock' || has(self.secretRef)",message="secretRef is required for api-key and oauth credential types"
```
gjkim42: nit: If we keep the bedrock type, there's a validation gap — a user can create a Task with `type: bedrock`, no `secretRef`, and no `region`. The controller would produce a job with only `CLAUDE_CODE_USE_BEDROCK=1` and nothing else. Would need something like:
```
self.type != 'bedrock' || has(self.secretRef) || (has(self.region) && self.region != '')
```
(Moot if we switch to `none`.)

internal/cli/run.go
```diff
@@ -108,10 +120,33 @@ func newRunCommand(cfg *ClientConfig) *cobra.Command {
...
+						if br.AccessKeyID == "" || br.SecretAccessKey == "" || br.Region == "" {
+							return fmt.Errorf("bedrock config requires accessKeyID, secretAccessKey, and region when using static credentials")
+						}
+						if !dryRun {
+							if err := ensureBedrockSecret(cfg, "kelos-credentials", br, yes); err != nil {
+								return err
+							}
+						}
+						secret = "kelos-credentials"
+					} else {
```
gjkim42: The bedrock CLI paths (both static and IRSA via config file, and via flags) have no dry-run tests in `dryrun_test.go`. The existing oauth/api-key paths are well-covered — would be good to match that coverage for whatever credential type we land on here.

gjkim42 (changes requested):
Thanks for the contribution! The `SecretRef` pointer change and the refactor to centralize credential injection are solid improvements.

My main concern is that `bedrock` as a credential type is too AWS-specific — see inline comments for a `none` credential type alternative that would make this provider-agnostic from day one.

api/v1alpha1/task_types.go
knechtionscoding: Love it

examples/09-bedrock-credentials/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Prerequisites
+
+1. Create an IAM role with Bedrock permissions
+2. Create a Kubernetes ServiceAccount annotated with the IAM role:
+
+   ```bash
+   kubectl create serviceaccount bedrock-agent-sa
+   kubectl annotate serviceaccount bedrock-agent-sa \
+     eks.amazonaws.com/role-arn=arn:aws:iam::123456789012:role/bedrock-agent-role
```
gjkim42, line 39: nit: The double colon in `arn:aws:iam::123456789012:role/...` looks like a typo — real ARNs have the account ID between the colons (`arn:aws:iam::<account-id>:role/...`). As-is it'll confuse copy-pasters. Consider `arn:aws:iam::111122223333:role/bedrock-agent-role` or adding a `<account-id>` placeholder.

knechtionscoding: @gjkim42 Apologies, but I'm not understanding this one. the example ARN provided is the correct format.

That is a real arn and matches the format you gave back in the comment.

gjkim42: I am not familiar with this. so you should be right.
Please squash commits and let me know.

PR #841, opened 2026-03-29
Add multi-architecture support for Docker images and workflows

### LINKED ISSUE #840: Add support for arm64
**AS A** User

**I WANT** to have multi-arch docker images

**SO THAT** I can run them on arm64 based machines as well

stearz:
#### What type of PR is this?
/kind feature

adds support for arm64
Fixes #840 

none
yes, containers now also run on arm64 machines

```release-note
Added support for arm64
```

---
## Summary by cubic
Add multi-architecture Docker builds (linux/amd64, linux/arm64) and update release workflows to push per-arch images and create multi-arch manifests, so images run on arm64 too (Fixes #840).
Makefile adds `IMAGE_PLATFORMS`, `PUSH`, and a `manifest` target; Dockerfiles now copy arch-specific binaries using `TARGETARCH`, and tag releases publish a `latest` multi-arch manifest.

<sup>Written for commit e6d86def2dd0b560102ff2c3677a5b947bfa518f. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -10,27 +10,11 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
-  release:
+  determine-version:
     runs-on: ubuntu-latest
-    permissions:
-      contents: write
-      packages: write
+    outputs:
+      version: ${{ steps.version.outputs.version }}
     steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-
-      - uses: actions/setup-go@v5
-        with:
-          go-version-file: go.mod
-
-      - name: Login to GHCR
-        uses: docker/login-action@v3
-        with:
-          registry: ghcr.io
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
       - name: Determine version
         id: version
         run: |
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -95,6 +99,27 @@ push: ## Push docker images (use WHAT to push specific image).
 		docker push $(REGISTRY)/$$(basename $$dir):$(VERSION); \
 	done
 
+IMAGE_PLATFORMS ?= linux/amd64,linux/arm64
+
+.PHONY: image-multiarch
+image-multiarch: ## Build and push multi-platform docker images via buildx (use WHAT to build specific image).
+	@for dir in $(filter cmd/%,$(or $(WHAT),$(IMAGE_DIRS))); do \
+		name=$$(basename $$dir); \
+		for arch in amd64 arm64; do \
```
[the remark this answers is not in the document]
stearz: fixed with 7300235

.github/workflows/release.yaml
```diff
@@ -24,6 +24,12 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Set up QEMU
```
knechtionscoding: My experience with this in #787 was that QEMU is very, very slow. like 60 minutes per run.

I swapped to running on ARM runners, example: https://github.com/datagravity-ai/kelos/blob/prod/.github/workflows/ci.yaml and it went to 2-3 minutes.

Makefile
```diff
@@ -95,6 +99,28 @@ push: ## Push docker images (use WHAT to push specific image).
 		docker push $(REGISTRY)/$$(basename $$dir):$(VERSION); \
 	done
 
+IMAGE_PLATFORMS ?= linux/amd64,linux/arm64
+IMAGE_ARCHES = $(shell echo "$(IMAGE_PLATFORMS)" | tr ',' '\n' | cut -d'/' -f2 | tr '\n' ' ')
+
+.PHONY: image-multiarch
+image-multiarch: ## Build and push multi-platform docker images via buildx (use WHAT to build specific image).
```
gjkim42: can we just use make image? I don't like to add two build rules for building the docker images.

.github/workflows/release.yaml
gjkim42: can we use a matrix strategy here for each architecture?

knechtionscoding: I updated: https://github.com/kelos-dev/kelos/pull/787 to use that. Feel free to steal that and close that one and merge this one @stearz

.github/workflows/release.yaml
```diff
@@ -10,70 +10,111 @@ concurrency:
...
 
-      - name: Push latest tags for releases
+      - name: Create and push latest manifests
         if: startsWith(github.ref, 'refs/tags/v')
-        run: |
-          make image VERSION=latest
-          make push VERSION=latest
+        env:
+          VERSION: ${{ needs.determine-version.outputs.version }}
+        run: make manifest IMAGE_PLATFORMS=linux/amd64,linux/arm64 VERSION=latest
```
gjkim42: P1: On tag pushes, the matrix build publishes `${VERSION}-amd64` and `${VERSION}-arm64`, but this step assembles `:latest` from `latest-amd64` and `latest-arm64`. Those source tags are never built in this workflow run, so a fresh release can fail here or repoint `:latest` at stale per-arch images from an older run. Please build and push `latest-$arch` before this step, or assemble `latest` from the versioned arch tags created above.

Makefile
```diff
@@ -1,7 +1,8 @@
 # Image configuration
 REGISTRY ?= ghcr.io/kelos-dev
 VERSION ?= latest
-IMAGE_DIRS ?= cmd/kelos-controller cmd/kelos-spawner cmd/kelos-token-refresher cmd/kelos-webhook-server cmd/ghproxy claude-code codex gemini opencode cursor
+IMAGE_DIRS ?= cmd/kelos-controller cmd/kelos-spawner cmd/kelos-token-refresher claude-code codex gemini opencode cursor
```
gjkim42: P1: Removing `cmd/ghproxy` and `cmd/kelos-webhook-server` from the default `IMAGE_DIRS` breaks existing callers of `make image`. The e2e workflow still runs `make image VERSION=e2e` and then loads `ghcr.io/kelos-dev/ghproxy:e2e`, so CI will fail because that tag is no longer built. It also stops publishing versioned `kelos-webhook-server` images even though the chart still references that image when the webhook server is enabled.

Makefile
```diff
@@ -79,14 +80,37 @@ build: ## Build binaries (use WHAT=cmd/kelos to build specific binary).
...
 	@for dir in $(or $(WHAT),$(IMAGE_DIRS)); do \
-		docker build -t $(REGISTRY)/$$(basename $$dir):$(VERSION) -f $$dir/Dockerfile .; \
+		docker buildx build --platform $(IMAGE_PLATFORMS) \
+			$(if $(filter true,$(PUSH)),--push,--load) \
+			-t $(REGISTRY)/$$(basename $$dir):$(VERSION) \
+			-f $$dir/Dockerfile .; \
+	done
+
+.PHONY: manifest
+manifest: ## Create and push multi-arch manifest from per-arch images (use WHAT, IMAGE_PLATFORMS).
```
gjkim42: Cleanup: after this refactor the release flow uses `make image ... PUSH=true` plus `make manifest`, and I could not find any remaining in-tree callers of `make push`. Keeping the old target around leaves a dead interface in the Makefile, so please remove `push` in the same cleanup.

stearz: cleaned up in e6d86de

Makefile
stearz: oops, forgot to get changes from upstream. Fixed in e6d86de

.github/workflows/release.yaml
stearz: fixed in e6d86de

Makefile
```diff
@@ -1,7 +1,8 @@
 # Image configuration
 REGISTRY ?= ghcr.io/kelos-dev
 VERSION ?= latest
-IMAGE_DIRS ?= cmd/kelos-controller cmd/kelos-spawner cmd/kelos-token-refresher cmd/kelos-webhook-server cmd/ghproxy claude-code codex gemini opencode cursor
+IMAGE_DIRS ?= cmd/kelos-controller cmd/kelos-spawner cmd/kelos-token-refresher cmd/ghproxy cmd/kelos-webhook-server claude-code codex gemini opencode cursor
```
gjkim42, line 4: **Bug:** `cmd/ghproxy` and `cmd/kelos-webhook-server` are back in `IMAGE_DIRS`, but their Dockerfiles weren't updated for multi-arch. The image build loop (line 89) renames binaries to `bin/$name-linux-$arch`, so `bin/ghproxy` no longer exists — only `bin/ghproxy-linux-amd64`. Their Dockerfiles still `COPY bin/ghproxy .` and `COPY bin/kelos-webhook-server .`, which will fail.

Needs the same `ARG TARGETARCH` + renamed COPY treatment as the other Dockerfiles:
```dockerfile
ARG TARGETARCH
COPY bin/ghproxy-linux-${TARGETARCH} ghproxy
```

Makefile
```diff
@@ -79,20 +80,39 @@ build: ## Build binaries (use WHAT=cmd/kelos to build specific binary).
...
 	go run ./cmd/kelos-controller
 
+IMAGE_PLATFORMS ?= linux/$(LOCAL_ARCH)
+IMAGE_ARCHES = $(shell echo "$(IMAGE_PLATFORMS)" | tr ',' '\n' | cut -d'/' -f2 | tr '\n' ' ')
+PUSH ?= false
+
 .PHONY: image
-image: ## Build docker images (use WHAT to build specific image).
+image: ## Build docker images (use WHAT, IMAGE_PLATFORMS, PUSH=true to customize).
 	@for dir in $(filter cmd/%,$(or $(WHAT),$(IMAGE_DIRS))); do \
```
gjkim42, line 89: **Nit:** `cmd/kelos-token-refresher` matches `cmd/%` here, so it gets pre-built and renamed to `bin/kelos-token-refresher-linux-$arch`. But its Dockerfile is a multi-stage build that rebuilds from source inside Docker — these pre-built binaries are never used. Not a correctness issue (the Docker build still works), but it wastes CI time. Consider either updating its Dockerfile to match the others, or excluding it from this loop.

PR #864, opened 2026-03-31
fix: reconciliation loop in taskspawner controller

gjkim42:
#### What type of PR is this?

/kind bug

The taskspawner controller was in an infinite reconciliation loop, constantly updating deployments and cronjobs even when nothing changed. This caused repeated pod restarts which exhausted the GitHub API rate limit, breaking all spawner discovery.

**Root cause:** `reflect.DeepEqual` was used to compare init containers and volumes between the desired spec (from the builder) and the live object (from the API server). Kubernetes normalizes objects by adding defaulted fields that the builder doesn't set:
- Init containers: `securityContext`, `terminationMessagePath`, `terminationMessagePolicy`
- Volumes: `Secret.DefaultMode` (defaults to 420)

This meant every reconciliation detected a "change" and called Update, which triggered another reconciliation.

**Fix:** Replace `reflect.DeepEqual` with semantic comparison functions (`initContainersEqual`, `volumesEqual`) that only compare the fields the builder explicitly sets — matching the pattern already used for main container comparison and `resourceRequirementsEqual`.

N/A

The comparison bug was pre-existing but was amplified by #859's new Task watcher, which added more reconciliation triggers. The fix applies to both `updateDeployment()` and `updateCronJob()`.

```release-note
NONE
```

---
## Summary by cubic
Fixes an infinite reconciliation loop in the taskspawner controller by using semantic comparisons for init containers and volumes. This stops unnecessary Deployment/CronJob updates, prevents pod restarts, and avoids GitHub API rate limit exhaustion.

- **Bug Fixes**
  - Compare only builder-set fields for init containers and volumes, ignoring Kubernetes defaults (securityContext, terminationMessage*, Secret.DefaultMode).
  - Treat empty init container `ImagePullPolicy` as matching any API-defaulted value to avoid false drift.
  - Apply the comparison in both Deployment and CronJob reconciliation paths.
  - Add tests to ensure defaults (including pull policy) are ignored and real changes are detected.

<sup>Written for commit a17fcdd4d62c6e3088d98b8dddea5f12441eecf0. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/taskspawner_controller.go
+++ b/internal/controller/taskspawner_controller.go
@@ -559,13 +559,13 @@ func (r *TaskSpawnerReconciler) updateDeployment(ctx context.Context, ts *kelosv
 	}
 
 	// Compare init containers (token-refresher sidecar)
-	if !reflect.DeepEqual(deploy.Spec.Template.Spec.InitContainers, desired.Spec.Template.Spec.InitContainers) {
+	if !initContainersEqual(deploy.Spec.Template.Spec.InitContainers, desired.Spec.Template.Spec.InitContainers) {
 		deploy.Spec.Template.Spec.InitContainers = desired.Spec.Template.Spec.InitContainers
 		needsUpdate = true
 	}
 
 	// Compare volumes (shared token emptyDir, github-app-secret)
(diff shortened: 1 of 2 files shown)


gjkim42 (commented):
/kelos review

PR #883, opened 2026-04-01
Migrate self-development to webhook-based TaskSpawners

gjkim42:
/kind cleanup

Two commits on top of `main`:

1. **Deploy GitHub webhook server in dev environment** (`.github/workflows/deploy-dev.yaml`) — enable `webhookServer.sources.github` via the new `kelos install -f` helm values flow so the webhook-based self-development TaskSpawners have a delivery endpoint in the dev cluster. A new preceding workflow step renders the full `github-webhook-secret` (HMAC `WEBHOOK_SECRET` + GitHub App `appID` / `installationID` / `privateKey`) from repo secrets and applies it via `kubectl apply -f -`, so the deploy is portable to another cluster without any pre-existing in-cluster secret. `githubSecretName` points at the same consolidated secret. Adds rollout restart + PodMonitoring for the `kelos-webhook-github` Deployment.

2. **Migrate self-development spawners to GitHub webhooks** (`self-development/*.yaml`, `self-development/README.md`, `internal/examples/self_development_test.go`) — replace the `githubIssues` / `githubPullRequests` polling sources in `kelos-workers`, `kelos-planner`, `kelos-reviewer`, `kelos-pr-responder`, `kelos-squash-commits`, and `kelos-triage` with `githubWebhook`. Cron-based spawners (fake-user, fake-strategist, config-update, self-update, image-update) are unchanged.

   Key design choices in the migration:
   - Lean on the webhook filter for state/label/draft gating instead of duplicating those checks in the prompts.
   - Drop the `/kelos needs-input` control command from the worker, planner, reviewer, and pr-responder prompts. It was a polling-era workaround to prevent re-pickup of stale state; webhook deliveries are discrete events and the bot's own replies don't match the trigger substrings, so the pause signal has no functional role. The `kelos/needs-input` label used by triage is a separate re-triage gate and stays.
   - Set `taskTemplate.branch: "{{.Branch}}"` on reviewer, pr-responder, and squash-commits so kelos checks out the real PR head and TaskReconciler can serialize concurrent writers via `task.spec.branch`.
   - `kelos-pr-responder` triggers on both `issue_comment` (with `/kelos pick-up`) and `pull_request_review` (submitted with `/kelos pick-up` in the review body) so a human review-change request that includes the trigger phrase resumes the loop.
   - `kelos-squash-commits` changes its completion marker to `Squash complete.` so the bot's own comment can't re-match the `/kelos squash-commits` trigger substring. The bot is also removed from the author filter as defense in depth.

   Adds `internal/examples/self_development_test.go` as a regression test that fails CI if any migrated spawner reverts to polling or drops the expected event/filter configuration.

N/A

Related but not fixed in this PR:
- #963 — self-development README is missing setup for GitHub App credentials inside the webhook secret. The deploy-dev workflow handles this via repo secrets, but users deploying manually still need docs.
- #964 — webhook server conditionally omits `.Branch` from template vars, breaking task creation with `missingkey=error`. Upstream fix required for the `branch: "{{.Branch}}"` idiom to work on fresh clusters without GitHub App credentials configured.

- Existing matching issues and PRs will not backfill after deploy. They need a fresh `/kelos pick-up`, `/kelos plan`, `/kelos review`, or relabel event to retrigger.
- The workflow depends on four new repository secrets: `KELOS_WEBHOOK_SECRET`, `KELOS_GITHUB_APP_ID`, `KELOS_GITHUB_APP_INSTALLATION_ID`, `KELOS_GITHUB_APP_PRIVATE_KEY`. These must be set before the next `deploy-dev` run.
- `webhookServer.gateway` is intentionally left at chart default. The dev cluster already has a Gateway (`kelos-webhook-gateway`) and HTTPRoute (`kelos-webhook-route`) applied out of band that rely on GKE Certificate Manager via the `networking.gke.io/certmap` annotation with no inline `tls.certificateRefs`, which the chart's gateway template cannot express. Managing them via helm would fight the working out-of-band config. There's a pointer comment in the values block to this effect.
- `task.spec.branch` is TaskReconciler's per-branch writer lock, which is why all PR-scoped spawners use `{{.Branch}}` rather than a synthetic name — concurrent spawners targeting the same PR head must share the lock.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/deploy-dev.yaml
+++ b/.github/workflows/deploy-dev.yaml
@@ -49,6 +49,30 @@ jobs:
           location: ${{ env.GKE_CLUSTER_LOCATION }}
           project_id: ${{ env.GCP_PROJECT_ID }}
 
+      - name: Apply github-webhook-secret
+        env:
+          KELOS_WEBHOOK_SECRET: ${{ secrets.KELOS_WEBHOOK_SECRET }}
+          KELOS_GITHUB_APP_ID: ${{ secrets.KELOS_GITHUB_APP_ID }}
+          KELOS_GITHUB_APP_INSTALLATION_ID: ${{ secrets.KELOS_GITHUB_APP_INSTALLATION_ID }}
+          KELOS_GITHUB_APP_PRIVATE_KEY: ${{ secrets.KELOS_GITHUB_APP_PRIVATE_KEY }}
+        run: |
+          # Apply the full github-webhook-secret from repo secrets so the
+          # workflow is self-sufficient and the secret state is portable to
+          # other clusters. Holds both the HMAC WEBHOOK_SECRET and the
+          # GitHub App credentials the webhook server uses to resolve
+          # .Branch for issue_comment events on pull requests.
+          KEY_FILE=$(mktemp)
+          trap 'rm -f "$KEY_FILE"' EXIT
+          printf '%s' "$KELOS_GITHUB_APP_PRIVATE_KEY" > "$KEY_FILE"
+          kubectl create secret generic github-webhook-secret \
+            --namespace kelos-system \
+            --from-literal=WEBHOOK_SECRET="$KELOS_WEBHOOK_SECRET" \
+            --from-literal=appID="$KELOS_GITHUB_APP_ID" \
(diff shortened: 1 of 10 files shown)


gjkim42:
@cubic-dev-ai

PR #866, opened 2026-03-31
feat: add Linear webhook support for TaskSpawners

knechtionscoding:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Adds Linear webhook support to the webhook server, allowing TaskSpawners to trigger task creation from Linear events (Issue, Comment, Project, etc.). This mirrors the existing GitHub webhook architecture and enables teams using Linear for project management to automatically spawn agent tasks from their Linear workflow.

Key changes:
- **New CRD fields**: `when.linearWebhook` with `types` and `filters` (action, states, labels, excludeLabels) filtering. Filters use OR semantics and action is validated via `+kubebuilder:validation:Enum`.
- **Webhook handler**: Extended `ServeHTTP` to accept Linear webhooks with HMAC-SHA256 signature validation. Linear dedup uses SHA-256 body hashing since Linear has no per-delivery ID header. Parsed event type (e.g., "Issue") is used in task naming instead of a generic "linear" prefix.
- **Filtering**: `MatchesLinearEvent` gates on `Types` first, then applies filters with OR semantics. Supports action, workflow state, required labels, and exclude labels with case-insensitive matching.
- **Helm chart**: Webhook server deployment, service, ingress, and Gateway API templates all support the `linear` source alongside GitHub. Toggled via `webhookServer.sources.linear.enabled`.
- **Controller**: CEL validation updated to require `workspaceRef` for `linearWebhook` source. Controller reconciles Linear webhook TaskSpawners with the same status/phase logic as GitHub.

#### Which issue(s) this PR is related to:

Fixes: #764
Fixes: #391

#### Special notes for your reviewer:

- Linear does not have an official Go SDK or typed webhook payload structs (unlike `go-github`). Webhook payloads are parsed manually from JSON in `ParseLinearWebhook`. This is consistent with how every other Go project handles Linear webhooks today.
- Linear supports the use of `Linear-Delivery` for deduplication so that is used, a fall back via sha of the entire webhook body is added
- Template variable documentation in the CRD was updated for both GitHub (added `{{.Repository}}`) and Linear sources.

#### Does this PR introduce a user-facing change?

```release-note
Add Linear webhook support for TaskSpawners. Configure `when.linearWebhook` with resource types and filters to automatically spawn tasks from Linear events (issues, comments, projects). Supports filtering by action (create/update/remove), workflow state, labels, and exclude labels.
```

---
## Summary by cubic
[...]

PR #898, opened 2026-04-03
fix: enrich Linear Comment webhook payloads with issue description

knechtionscoding:
#### What type of PR is this?

/kind bug

Linear Comment webhooks include `data.issue.id`, `data.issue.identifier`, `data.issue.title`, and `data.issue.url` — but **not** `data.issue.description`. This causes Go template rendering to crash with `map has no entry for key "description"` when spawner prompt templates reference `{{.Payload.data.issue.description}}`.

This fix extends the existing label enrichment (from PR #866 ) to also fetch the issue description from the Linear GraphQL API. Both fields are now fetched in a single API call, and enrichment is unconditional for all Comment events — preventing a class of template errors cheaply.

**Changes:**
- Extended the Linear GraphQL query to fetch `description` alongside `labels`
- Made Comment event enrichment unconditional (moved before the spawner loop, no longer gated on `spawnerNeedsLinearLabels`)
- Description is only injected if not already present in the payload (preserves existing values)
- Removed now-unused `spawnerNeedsLinearLabels` function

N/A

The enrichment is now unconditional for Comment events. This means a Linear API call happens for every Comment webhook even if no spawner uses label filtering. This is intentional — it's cheap (single GraphQL call) and prevents template crashes.

```release-note
Fixed template rendering crash for Linear Comment webhooks by enriching payloads with issue description from the Linear API.
```

---
## Summary by cubic
Fixes template crashes for Linear Comment webhooks by enriching payloads with the issue description (and labels) via a single GraphQL call before filtering. All Comment events now get consistent data to avoid missing fields in templates.

- **Bug Fixes**
  - Fetch issue description and labels in one GraphQL call and inject into `data.issue`; only add description if absent.
  - Run enrichment unconditionally for Comment events before spawner filters to prevent label-based mismatches and template errors.
  - Introduced `fetchLinearIssueDetails`, updated tests, and removed the unused `spawnerNeedsLinearLabels`.

<sup>Written for commit 09bcfb9d2819bb8c03884ef8375572b86ac49b7c. Summary will update on new commits.</sup>

### DIFF
--- a/internal/webhook/handler.go
+++ b/internal/webhook/handler.go
@@ -280,6 +280,15 @@ func (h *WebhookHandler) processWebhook(ctx context.Context, eventType string, p
 
 	log.Info("Processing webhook event", "resourceID", parsed.ID, "title", parsed.Title)
 
+	// Unconditionally enrich Linear Comment events with issue details
+	// (labels + description) before evaluating spawner filters. Linear does
+	// not include these fields in Comment webhook payloads, which causes
+	// label filtering to miss and template rendering to crash. This is a
+	// single cheap GraphQL call per delivery.
+	if parsed.Linear != nil {
+		enrichLinearCommentIssue(ctx, log, parsed.Linear)
+	}
+
 	// Get all TaskSpawners that match this source type
 	spawners, err := h.getMatchingSpawners(ctx)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


knechtionscoding:
> I think I don't want to expand an extra field, since it's too specific and we have an alternative way to query it inside of the task.

@gjkim42 I agree with avoiding unnecessarily expanding, but it feels like there should be parity when writing the task spawner, otherwise one has to be very defensive in writing task spawners (and duplicate a lot of code/references). If I want to have the same taskrespond to both labeling and comments, for example, then I have to write a very defensive task spawner. That's okay, I just figured I'd expand on the tradeoff and why this was a follow on

PR #902, opened 2026-04-03
feat(api): Add generic webhook source for arbitrary event-driven task spawning

### LINKED ISSUE #687: Integration: Add generic webhook source type to TaskSpawner for universal event-driven task triggering

## Summary

Kelos currently supports four source types: `githubIssues`, `githubPullRequests`, `cron`, and `jira`. Each is a purpose-built integration with its own discovery logic. Adding new sources (Slack [#595], Alertmanager [#570], Linear [#391], Sentry, PagerDuty, Datadog, etc.) requires implementing a new `Source` interface, extending the `When` struct, updating the spawner binary, and releasing a new version — for every single integration.

This proposal adds a **generic `webhook` source type** that accepts arbitrary HTTP POST payloads and maps them to `WorkItem` fields via configurable JSONPath expressions. This enables integration with **any** webhook-capable system without Kelos-side code changes.

This is distinct from #268, which proposes GitHub-specific webhook handling (GitHub event types, `X-Hub-Signature-256` validation, GitHub payload structure). The generic webhook source proposed here handles **arbitrary JSON payloads** from any HTTP caller.

### 1. Integration bottleneck

Every new source requires a full development cycle:
- New types in `api/v1alpha1/taskspawner_types.go` (new field on `When`)
- New source implementation in `internal/source/`
- New `buildSource` branch in `cmd/kelos-spawner/main.go:461-538`
- New deployment builder logic in `internal/controller/taskspawner_deployment_builder.go`
- CRD regeneration, tests, docs

There are currently **6 open integration proposals** (#595 Slack, #570 Alertmanager, #391 Linear, #603 githubCheckRuns, #636 githubSecurityAlerts, #373 GitLab) — each requiring this full cycle. And that doesn't cover Sentry, PagerDuty, Datadog, Argo CD, Terraform Cloud, or internal tools.

### 2. No escape hatch for custom workflows

Teams with internal tools, custom CI systems, or niche platforms have no way to trigger Kelos tasks from their systems. They must either:
- Build an external controller that watches their system and creates Task resources directly (bypasses TaskSpawner governance: maxConcurrency, maxTotalTasks, suspend, reporting, dedup)
- Wait for Kelos to add support for their specific platform

### 3. Polling latency for all sources

All current sources are poll-based (the spawner runs a `Discover` loop on `pollInterval`). Even with a 1-minute poll interval, there's inherent latency. A generic webhook source would enable real-time, push-based triggering.

## Proposed API

Add a new `Webhook` field to the `When` struct:

```yaml
kind: TaskSpawner
metadata:
  name: sentry-error-fixer
spec:
  when:
    webhook:
      # Authentication for incoming requests
      auth:
        # HMAC signature validation (e.g., Sentry, GitHub, Stripe)
        hmac:
          secretRef:
            name: sentry-webhook-secret
          header: "Sentry-Hook-Signature"  # HTTP header containing signature
          algorithm: sha256                # hmac-sha256 (default), hmac-sha1

      # Map JSON payload fields to WorkItem template variables
      # Uses JSONPath expressions evaluated against the request body
      fieldMapping:
        id: "$.event.event_id"
        title: "$.event.title"
        body: "$.event.metadata.value"
        url: "$.url"
        # Nested field extraction
        labels: "$.event.tags[*].key"

      # Optional: only trigger when payload matches these conditions
      filters:
        - field: "$.event.level"
          value: "error"              # exact match
        - field: "$.event.platform"
          pattern: "python|go|node"   # regex match

      # Port for the webhook receiver (default: 8080)
      port: 8080

  taskTemplate:
    type: claude-code
    workspaceRef:
      name: my-app
    credentials:
      type: oauth
      secretRef:
        name: claude-creds
    promptTemplate: |
      A production error was detected:

      **Error**: {{.Title}}
      **Details**: {{.Body}}
      **Error URL**: {{.URL}}

      Investigate this error in the codebase:
      1. Find the relevant source file from the stack trace
      2. Understand the root cause
      3. Create a fix and open a PR
      4. Add a regression test
```

### Go types

```go
// Webhook configures a generic HTTP webhook receiver for event-driven task spawning.
// Any system that can send an HTTP POST with a JSON body can trigger tasks.
type Webhook struct {
    // Auth configures authentication for incoming webhook requests.
    // +optional
    Auth *WebhookAuth `json:"auth,omitempty"`

    // FieldMapping maps JSONPath expressions to WorkItem fields.
    // Each key is a WorkItem field name (id, title, body, url, labels),
    // and each value is a JSONPath expression evaluated against the
    // request body. The "id" field is required for deduplication.
    FieldMapping map[string]string `json:"fieldMapping"`

    // Filters define conditions that must all match for a webhook
    // delivery to trigger a task. If empty, all deliveries trigger tasks.
    // +optional
    Filters []WebhookFilter `json:"filters,omitempty"`

    // Port is the port the webhook receiver listens on. Defaults to 8080.
    // +kubebuilder:default=8080
    // +optional
    Port int32 `json:"port,omitempty"`

    // Path is the URL path the webhook receiver listens on. Defaults to "/".
    // +kubebuilder:default="/"
    // +optional
    Path string `json:"path,omitempty"`
}

type WebhookAuth struct {
    // HMAC configures HMAC signature validation for incoming requests.
    // +optional
    HMAC *HMACAuth `json:"hmac,omitempty"`

    // BearerToken configures Bearer token validation.
    // +optional
    BearerToken *BearerTokenAuth `json:"bearerToken,omitempty"`
}

type HMACAuth struct {
    // SecretRef references a Secret containing the HMAC shared secret
    // (key: "WEBHOOK_SECRET").
    SecretRef SecretReference `json:"secretRef"`

    // Header is the HTTP header containing the HMAC signature.
    // Defaults to "X-Hub-Signature-256".
    // +kubebuilder:default="X-Hub-Signature-256"
    // +optional
    Header string `json:"header,omitempty"`

    // Algorithm is the HMAC algorithm. Defaults to "sha256".
    // +kubebuilder:validation:Enum=sha256;sha1
    // +kubebuilder:default=sha256
    // +optional
    Algorithm string `json:"algorithm,omitempty"`
}

type BearerTokenAuth struct {
    // SecretRef references a Secret containing the expected Bearer token
    // (key: "WEBHOOK_TOKEN").
    SecretRef SecretReference `json:"secretRef"`
}

type WebhookFilter struct {
    // Field is a JSONPath expression selecting the payload field to match.
    Field string `json:"field"`

    // Value requires an exact match against the extracted field value.
    // Mutually exclusive with Pattern.
    // +optional
    Value string `json:"value,omitempty"`

    // Pattern requires a regex match against the extracted field value.
    // Mutually exclusive with Value.
    // +optional
    Pattern string `json:"pattern,omitempty"`
}
```

### Template variables

Webhook-sourced WorkItems expose all mapped fields through the standard template variables (`{{.ID}}`, `{{.Title}}`, `{{.Body}}`, `{{.URL}}`, `{{.Labels}}`), plus a new catch-all:

- `{{.Payload}}` — the full raw JSON body (for prompt injection of arbitrary context)

### Spawner modifications

The spawner currently operates in two modes:
- **Poll loop** (Deployment): for `githubIssues`, `githubPullRequests`, `jira`
- **One-shot** (CronJob): for `cron`

The webhook source adds a third mode:
- **HTTP server** (Deployment + Service): for `webhook`

When `webhook` is configured, the spawner starts an HTTP server instead of a poll loop. On each validated request, it converts the payload to a `WorkItem` via `fieldMapping` and runs the standard task-creation logic (`runCycleWithSource` with a single-item list).

```
┌──────────────┐    HTTP POST     ┌──────────────────┐
│  External    │ ───────────────▶ │  Spawner Pod     │
│  System      │                  │  (HTTP server)   │
│  (Sentry,    │                  │                  │
│   PagerDuty, │                  │  ┌────────────┐  │
│   custom)    │                  │  │ Validate   │  │
└──────────────┘                  │  │ auth       │  │
                                  │  └─────┬──────┘  │
                                  │        │         │
                                  │  ┌─────▼──────┐  │
                                  │  │ Extract    │  │
                                  │  │ fields via │  │
                                  │  │ JSONPath   │  │
                                  │  └─────┬──────┘  │
                                  │        │         │
                                  │  ┌─────▼──────┐  │
                                  │  │ Apply      │  │
                                  │  │ filters    │  │
                                  │  └─────┬──────┘  │
                                  │        │         │
                                  │  ┌─────▼──────┐  │
                                  │  │ Create     │  │
                                  │  │ Task       │  │
                                  │  └────────────┘  │
                                  └──────────────────┘
```

### Controller modifications

The TaskSpawner controller (`internal/controller/taskspawner_controller.go`) needs to:

1. Create a `Service` resource alongside the Deployment when `webhook` is configured
2. Optionally create an `Ingress` if the user opts in (or document that users should create their own Ingress/LoadBalancer)

The deployment builder (`internal/controller/taskspawner_deployment_builder.go`) adds a `--webhook-mode` flag and port configuration to the spawner container.

### Deduplication

Uses the standard `<spawner>-<item.ID>` pattern. The `id` field in `fieldMapping` is required — it must map to a unique identifier in the webhook payload (e.g., event ID, alert ID, incident number). This ensures:
- Retried webhook deliveries don't create duplicate tasks
- The standard `maxConcurrency` and `maxTotalTasks` governance applies

### 1. Sentry: Auto-fix production errors

```yaml
kind: TaskSpawner
metadata:
  name: sentry-auto-fix
spec:
  when:
    webhook:
      auth:
        hmac:
          secretRef:
            name: sentry-secret
          header: "Sentry-Hook-Signature"
      fieldMapping:
        id: "$.data.event.event_id"
        title: "$.data.event.title"
        body: "$.data.event.exception.values[0].value"
        url: "$.data.url"
      filters:
        - field: "$.action"
          value: "created"
        - field: "$.data.event.level"
          value: "error"
  maxConcurrency: 2
  taskTemplate:
    type: claude-code
    workspaceRef:
      name: my-app
    credentials:
      type: oauth
      secretRef:
        name: claude-creds
    promptTemplate: |
      Production error detected: {{.Title}}
      Details: {{.Body}}
      Sentry URL: {{.URL}}

      Find and fix this error, then open a PR with a regression test.
```

### 2. PagerDuty: Incident-driven remediation

```yaml
kind: TaskSpawner
metadata:
  name: pagerduty-auto-remediate
spec:
  when:
    webhook:
      auth:
        hmac:
          secretRef:
            name: pagerduty-secret
          header: "X-PagerDuty-Signature"
      fieldMapping:
        id: "$.messages[0].incident.id"
        title: "$.messages[0].incident.title"
        body: "$.messages[0].incident.description"
        url: "$.messages[0].incident.html_url"
      filters:
        - field: "$.messages[0].event"
          value: "incident.trigger"
  maxConcurrency: 1
  taskTemplate:
    type: claude-code
    workspaceRef:
      name: infra-repo
    credentials:
      type: oauth
      secretRef:
        name: claude-creds
    promptTemplate: |
      Incident triggered: {{.Title}}
      Description: {{.Body}}
      PagerDuty URL: {{.URL}}

      Investigate this incident and create a remediation PR.
```

### 3. Custom CI: Post-deploy verification

```yaml
kind: TaskSpawner
metadata:
  name: post-deploy-verify
spec:
  when:
    webhook:
      auth:
        bearerToken:
          secretRef:
            name: ci-webhook-token
      fieldMapping:
        id: "$.deployment_id"
        title: "$.service_name"
        body: "$.changelog"
      filters:
        - field: "$.environment"
          value: "production"
  maxConcurrency: 1
  taskTemplate:
    type: claude-code
    workspaceRef:
      name: my-app
    credentials:
      type: api-key
      secretRef:
        name: anthropic-key
    promptTemplate: |
      Service {{.Title}} was just deployed to production.
      Changes: {{.Body}}

      Run smoke tests and verify the deployment looks healthy.
      If you find issues, create a rollback PR.
```

## Relationship to other proposals

| Proposal | Relationship |
|----------|-------------|
| #268 (GitHub webhooks) | Complementary. #268 is a GitHub-optimized source with deep event-type awareness. This proposal is a generic fallback for non-GitHub systems. Both could coexist — use `githubWebhook` for GitHub, `webhook` for everything else. |
| #595 (Slack), #570 (Alertmanager), #391 (Linear) | Partially subsumed. These systems all support outgoing webhooks. A generic webhook source handles them without Kelos needing to know their APIs. However, dedicated sources could still be valuable for bi-directional integration (e.g., Slack interactive messages). |
| #603 (githubCheckRuns), #636 (githubSecurityAlerts) | Not subsumed. These require GitHub API polling for specific data types that don't have direct webhook equivalents with full context. |

### Phase 1: Core webhook receiver (smallest viable increment)
- Add `Webhook` to `When` struct in `api/v1alpha1/taskspawner_types.go`
- Implement `WebhookSource` in `internal/source/webhook.go` with JSONPath field extraction
- Add HTTP server mode to `cmd/kelos-spawner/main.go`
- Controller creates `Service` for webhook-mode spawners
- HMAC auth support
- **Scope**: ~400 lines + tests

### Phase 2: Filters and extended auth
- Payload filtering with exact match and regex
- Bearer token auth
- Response body configuration (for systems that expect specific responses)

### Phase 3: Quality of life
- `kelos get taskspawner` shows webhook URL/endpoint
- Example configurations for common platforms (Sentry, PagerDuty, Datadog)
- Optional Ingress resource creation

## Backward compatibility

- Adding `Webhook *Webhook` to `When` is a non-breaking additive change
- Existing spawners using `githubIssues`, `githubPullRequests`, `cron`, or `jira` are unaffected
- XValidation rule on `When` needs updating to include `webhook` in the mutual exclusivity check
- The `Source` interface remains unchanged — the webhook receiver wraps incoming payloads into `WorkItem` and reuses the existing task creation logic

### Why JSONPath over CEL?

JSONPath is simpler, more widely understood, and sufficient for field extraction. CEL (#498) is better suited for complex boolean predicates. They serve different purposes and can coexist. The `filters` field could adopt CEL later if #498 is implemented.

### Why not a generic adapter pattern instead?

An alternative is a "webhook adapter" sidecar that receives webhooks and calls the Kubernetes API to create Tasks directly. This bypasses TaskSpawner governance (maxConcurrency, dedup, suspend) and requires users to manage RBAC. Keeping webhook handling inside the spawner preserves all existing governance features.

### Why a Deployment+Service (not an Ingress)?

The spawner creates a ClusterIP Service, which is sufficient for in-cluster callers. For external callers, teams have diverse ingress strategies (Nginx, Istio, cloud LBs) — Kelos should not opine on this. Documentation should guide users to expose the Service using their existing ingress infrastructure.

## References

- Current `When` struct: `api/v1alpha1/taskspawner_types.go:22-39`
- Source interface: `internal/source/source.go:37-39`
- Spawner `buildSource`: `cmd/kelos-spawner/main.go:461-538`
- TaskSpawner controller: `internal/controller/taskspawner_controller.go`
- Deployment builder: `internal/controller/taskspawner_deployment_builder.go`
- WorkItem struct: `internal/source/source.go:10-33`
- Related: #268 (GitHub webhooks — complementary, GitHub-specific)
- Related: #595, #570, #391 (specific integrations that could use generic webhook)

knechtionscoding:
#### What type of PR is this?

/kind feature

Adds a generic webhook source type (`spec.when.webhook`) to TaskSpawner that accepts arbitrary HTTP POST payloads with JSON bodies and maps them to task template variables via configurable JSONPath expressions. This enables integration with any webhook-capable system (Notion, Sentry, Drata, Salesforce, Aikido, etc.) without requiring Kelos-side code changes for each new integration.

Key design decisions:
- **Path-based routing**: Each source gets its own path (`/webhook/<source>`) and derives its HMAC secret from the `<SOURCE>_WEBHOOK_SECRET` environment variable (e.g., `/webhook/notion` → `NOTION_WEBHOOK_SECRET`)
- **JSONPath field extraction**: `fieldMapping` maps JSONPath expressions to template variables available in `promptTemplate` and `branch`
- **AND-semantics filters**: Filters support exact match and regex, all must match for a delivery to trigger a task
- **Plugs into existing infrastructure**: Reuses the `kelos-webhook-server` with `--source generic` rather than creating per-spawner Deployments
- **No workspaceRef required**: Generic webhook spawners can optionally omit `workspaceRef` since they may not be git-repo-driven
- **Helm chart support**: Adds `webhookServer.sources.generic` with `envFrom`-based multi-secret injection so all per-source secrets come from a single Secret resource

Fixes #687

- The generic source uses `envFrom` in the helm chart (not `valueFrom`) since users will have multiple `<SOURCE>_WEBHOOK_SECRET` keys in one Secret
- The `SignaturePrefix` field defaults to `"sha256="` but can be set to `""` for providers that send raw hex digests (e.g., Linear-style)
- Filter semantics are AND (all must match), unlike GitHub/Linear webhook filters which use OR. AND is more intuitive for generic payloads since there are no event-type scopes. OR can be achieved by creating multiple TaskSpawners
- The ingress/gateway routes use `/webhook/` as a catch-all prefix for generic, which is placed after the more specific `/webhook/github` and `/webhook/linear` rules
- Added `github.com/PaesslerAG/jsonpath` dependency for JSONPath evaluation

```release-note
feat: Add generic webhook source type for TaskSpawner that accepts arbitrary HTTP POST payloads and maps JSON fields to task template variables via JSONPath expressions, enabling integration with any webhook-capable system without Kelos-side code changes.
```

---
## Summary by cubic
Adds a generic `webhook` source to TaskSpawner to spawn tasks from any JSON POST at `/webhook/<source>` using JSONPath field mapping and AND-filters, with id-based idempotency when `fieldMapping.id` is set (fallback to body‑hash). Implements #687.

- **New Features**
  - Generic webhooks: `POST /webhook/<source>` routes to the generic handler; delivery IDs use `fieldMapping.id` when present, else body‑hash.
  - JSONPath mapping: `fieldMapping` populates template vars (`id` required); `ID`, `Title`, `Body`, `URL` default to empty; `.Payload` always available.
  - Filters: AND semantics with exact or regex matches; spawners match on `source`.
  - Server/Charts: `kelos-webhook-server --source=generic` serves under `/webhook/`; adds generic Deployment/Service and `/webhook/` routes to ingress/gateway; controller treats generic as webhook‑driven; CRD validations added.
  - Dependencies: add `github.com/PaesslerAG/jsonpath`.

- **Migration**
  - Helm: enable `webhookServer.sources.generic.enabled`, optionally set `webhookServer.sources.generic.secretName` for env injection.
  - Define a TaskSpawner with `spec.when.webhook` (`source`, `fieldMapping` with `id`, optional `filters`), then send POSTs to `/webhook/<source>`.

<sup>Written for commit f96043247d423c8f1568e0f28697195f50dfa571. Summary will update on new commits.</sup>


api/v1alpha1/taskspawner_types.go
```diff
@@ -424,6 +431,78 @@ type LinearWebhookFilter struct {
...
+// +kubebuilder:validation:XValidation:rule="'id' in self.fieldMapping",message="fieldMapping must include an 'id' key for deduplication and task naming"
+type GenericWebhook struct {
+	// Source is a short identifier for this webhook source (e.g., "notion",
+	// "sentry", "drata"). It determines:
+	//   - The URL path: /webhook/<source>
+	//   - The env var for HMAC validation: <SOURCE>_WEBHOOK_SECRET
+	// Must be lowercase alphanumeric with optional hyphens.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`
+	Source string `json:"source"`
```
gjkim42, line 448: The source should be unique in cluster scope, right?

knechtionscoding:
@gjkim42 bump :)

knechtionscoding:
@gjkim42 updated so the fork e2e runs

knechtionscoding:
To be clear, we've been using this internally and this works perfectly fine. We haven't tested at scale yet (only a few webhook inbound sources right now) but it's running smoothly for everything.

We aren't yet sure how much we want to use it vs funneling everything through linear. But I think that's a general tactical thing and generic webhooks allows support for more tools without having to expand the API surface

gjkim42:
I am generally positive about the feature. My only concern is API is hard to fix, so I'd like to find the best way to design it.

api/v1alpha1/taskspawner_types.go
```diff
@@ -44,6 +44,13 @@ type When struct {
 	// LinearWebhook triggers task spawning on Linear webhook events.
 	// +optional
 	LinearWebhook *LinearWebhook `json:"linearWebhook,omitempty"`
+
+	// Webhook triggers task spawning from arbitrary HTTP POST payloads.
+	// Any system that can send an HTTP POST with a JSON body can trigger
+	// tasks through this source. The URL path is /webhook/<source> and
+	// the HMAC secret is read from the <SOURCE>_WEBHOOK_SECRET env var.
+	// +optional
+	Webhook *GenericWebhook `json:"webhook,omitempty"`
```
gjkim42: can we name it GenericWebhook here?

api/v1alpha1/taskspawner_types.go
```diff
@@ -424,6 +431,78 @@ type LinearWebhookFilter struct {
...
+	// +optional
+	SignatureHeader string `json:"signatureHeader,omitempty"`
+
+	// SignaturePrefix is the prefix before the hex digest in the signature
+	// header value (e.g., "sha256=" for GitHub-style signatures). Set to
+	// empty string if the header contains the raw hex digest (Linear-style).
+	// Defaults to "sha256=".
+	// +kubebuilder:default="sha256="
+	// +optional
+	SignaturePrefix string `json:"signaturePrefix,omitempty"`
```
gjkim42: I am not sure what these 3 fields are about. Can you elaborate on them?
Are they for authentication?

knechtionscoding: Validation of the incoming webhook.

Example: [Github](https://docs.github.com/en/webhooks/using-webhooks/validating-webhook-deliveries) where `x-hub-signature-256` is provided.

[Linear](https://linear.app/developers/webhooks#webhook-payload) for example provides the following:

1. Linear-Delivery which is an ID for idempotency
2. Linear-Signature which is the signature header.

Other examples include [notion](https://developers.notion.com/reference/webhooks#step-3-validating-event-payloads-recommended) (which uses a prefix for the sha unlike linear), [salesforce](https://developer.salesforce.com/docs/sales/spiff/guide/webhook-api-get-started.html) who uses just a `signature` header, or [gitlab](https://docs.gitlab.com/user/project/integrations/webhooks/#delivery-headers) who provides X-Gitlab-Event-UUID, or Idempotency-Key, and X-Gitlab-Token for idempotency and validation.

Because there is slight variation in the format of the signature header then these three fields are required.

DeliveryIDHeader to allow for definition of the idempotency header (if supported)
SignatureHeader to allow for definition of the signature header (if desired)
SignaturePrefix to allow for the prefix to strip (if the header looks like github or salesforce where it is prefixed with sha256= vs linear who just uses the sha directly).

api/v1alpha1/taskspawner_types.go
knechtionscoding: Yes. The source is unique in the cluster, but we aren't spawning a new webhook server per source (at least not currently).

The current implementation spawns a single GenericWebhook taskSpawner that then listens for everything that isn't routed to one of our specific handlers. This means on the taskSpawner you have to declare which source you care to listen too. 

For example, if you are listening to both notion and salesforce on the taskspawners for those you need to declare what source you care about.

api/v1alpha1/taskspawner_types.go
knechtionscoding: Updated

knechtionscoding:
@gjkim42 Updated based on review comments and ensured verify passes.

api/v1alpha1/taskspawner_types.go
gjkim42: I am not sure if this is appropriate to be inside the taskspawner, not the webhook.
Can we defer these fields for now? What do you think?

knechtionscoding: I think we can defer them. Are you planning on having a separate webhook CRD where these can be declared?

gjkim42: I think we need something similar with it. maybe webhook or webhookGateway...?
Do you think current webhook model is better than making your own custom webhook server?
(I just want to get some feedback on this single native webhook model)

knechtionscoding: >I think we need something similar with it. maybe webhook or webhookGateway...?

I could see a webhookGateway/webhook being useful to be able to declare these. This would also let us consolidate github and linear into a single model. There's nothing unique about those (as long as we allow for context to be gathered in some manner before we template out the task spawner). 

> Do you think current webhook model is better than making your own custom webhook server?

I do think this is better than a custom webhook server. A custom webhook server can still be made but this method allows for a lot lower friction in adoption. Having a generic one allows for you to install kelos into the cluster and start sending webhooks to it, rather than having to write a orchestration layer.

gjkim42: thank you.
Let's make the minimal API for this PR.

knechtionscoding: Sure!

api/v1alpha1/taskspawner_types.go
knechtionscoding: @gjkim42 Any concerns here? Otherwise I'll rework the other thread and push for final review

api/v1alpha1/taskspawner_types.go
knechtionscoding: Updated

api/v1alpha1/taskspawner_types.go
gjkim42: Is there a validation that prevents duplicate sources?

knechtionscoding: No, because it isn't declared anywhere as a universal.

You are allowed to have multiple task spawners listening to the same source.

And on the webhook server side it just sees the path and reads for the secret. There's no ability for duplication anywhere.

Because once deployed you can send to webhook/test, webhook/notion, webhook/drata without any further configuration on the server side.

Once the generic source is deployed it then just checks for an existing taskSpawner with source that matches what it received.

So there's nothing to ensure isn't duplicated.

internal/webhook/handler.go
```diff
@@ -198,6 +206,26 @@ func (h *WebhookHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+		eventType = sourceName
+
+		// Single API list call provides matching spawners, avoiding a
+		// redundant list in processWebhook.
+		genericSpawners = h.getGenericSpawners(ctx)
+
+		// Derive delivery ID from body hash
+		sum := sha256.Sum256(body)
+		deliveryID = "generic-" + sourceName + "-" + hex.EncodeToString(sum[:])
```
gjkim42: P2: dedupe here is driven by a hash of the raw request body, not the required fieldMapping.id. That means a provider can retry the same logical resource with the same stable mapped id but slightly different JSON encoding or extra metadata and bypass deliveryCache, creating a second Task.

internal/webhook/generic_filter.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// ExtractFields evaluates JSONPath expressions from fieldMapping against
+// the payload and populates the Fields map. Missing fields produce empty
+// strings rather than errors so that optional mappings do not block task
+// creation.
+func (e *GenericEventData) ExtractFields(fieldMapping map[string]string) error {
+	e.Fields = make(map[string]string, len(fieldMapping))
+	for key, expr := range fieldMapping {
+		val, err := jsonpath.Get(expr, e.Payload)
```
gjkim42, line 65: P2: this treats every jsonpath.Get error as a missing field, which also swallows malformed JSONPath expressions in required mappings like id or title. With a typo such as $.data[, the webhook proceeds with empty strings instead of surfacing a configuration error, which makes broken generic webhook specs hard to diagnose and can produce blank task metadata.

internal/webhook/handler.go
knechtionscoding: Done

internal/webhook/generic_filter.go
knechtionscoding: Done

gjkim42 (approved):

PR #931, opened 2026-04-08
feat(webhooks): parse out commenter relevant fields when receiving a github webhook

knechtionscoding:
#### What type of PR is this?

/kind feature

Extracts comment-specific fields (CommentBody, CommentURL) from GitHub webhook payloads for issue_comment, pull_request_review, and pull_request_review_comment events and exposes them as template variables for TaskSpawner prompt templates.

Previously, these fields were never parsed or included in the template variables map, causing prompt template execution to fail when a spawner referenced comment-specific variables.

CommentAuthor was intentionally omitted because the existing Sender field already captures the same user who triggered the webhook event.

N/A

The Sender field (which was already populated) represents the user who triggered the webhook event, which is the same as the comment author. CommentBody and CommentURL provide access to the comment content and link, which are distinct from the parent issue/PR Body and URL.

Also updates the template variable tables in docs/reference.md and docs/integration.md to document the new variables.

```release-note
Add CommentBody and CommentURL template variables for GitHub comment webhook events (issue_comment, pull_request_review, pull_request_review_comment).
```

### DIFF
--- a/internal/webhook/github_filter.go
+++ b/internal/webhook/github_filter.go
@@ -35,6 +35,10 @@ type GitHubEventData struct {
 	Body   string
 	URL    string
 	Branch string
+	// Comment-specific fields for issue_comment, pull_request_review,
+	// and pull_request_review_comment events.
+	CommentBody string
+	CommentURL  string
 	// ChangedFiles lists file paths modified by the event.
 	// For push events, populated from the payload. For PR events, lazily
 	// fetched from the GitHub API when a webhook filter uses FilePatterns.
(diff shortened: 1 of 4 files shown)


knechtionscoding:
@gjkim42 I think this is ready for review

knechtionscoding:
bump @gjkim42

knechtionscoding:
@gjkim42 Bump again now that you are back :)

internal/webhook/github_filter.go
```diff
@@ -35,6 +35,11 @@ type GitHubEventData struct {
 	Body   string
 	URL    string
 	Branch string
+	// Comment-specific fields for issue_comment, pull_request_review,
+	// and pull_request_review_comment events.
+	CommentAuthor string
+	CommentBody   string
+	CommentURL    string
```
gjkim42: Do we need a concrete use case for this?
Is there an alternative for this?

I am hesitant to add another event-specific variable.
Also not sure about the field name...  do we have to use Sender, Body, and URL instead of CommentAuthor, CommentBody, and CommentURL?

knechtionscoding: @gjkim42 So, we can drop CommentAuthor (that is in practice always the same as sender) but body and url are distinct from CommentBody and CommentURL.

Specifically te ability to support filtering/templating based solely on the comment body is relevant has a usecase.

knechtionscoding: And specifically being able to respond to individual comments is useful, for example, having kelos solve individual feedbacks/process ideas.

knechtionscoding:
@gjkim42 feedback resolved

gjkim42 (approved):
Thanks!

PR #1010, opened 2026-04-18
fix: validate that --prompt is not empty in axon run

### LINKED ISSUE #199: CLI UX: 'axon run' should validate that --prompt is not empty
## Problem

The \`axon run\` command accepts an empty string for the required \`--prompt\` flag, creating invalid Tasks that will fail at the API level.

### Current Behavior

```bash
# Empty prompt is accepted
axon run -p "" --secret dummy --dry-run
# Output:
apiVersion: axon.io/v1alpha1
kind: Task
metadata:
  name: task-twq97
spec:
  prompt: ""  # Empty!
  ...
```

If submitted to the cluster (without \`--dry-run\`), this would fail at the Kubernetes API validation level, but only after:
1. The task resource is attempted to be created
2. Network round-trip to the API server
3. API-level validation runs

### Why This is a Problem

1. **Late failure**: Error happens at API submission, not immediately
2. **Poor error location**: Users don't get helpful context about which CLI flag is the problem
3. **Wasted resources**: Unnecessary API call for an obvious client-side issue
4. **Incomplete client validation**: The CLI marks \`--prompt\` as required (\`cmd.MarkFlagRequired("prompt")\`) but doesn't validate it's non-empty

### Expected Behavior

The CLI should validate that the prompt is not empty or whitespace-only:

```bash
axon run -p "" --secret dummy
Error: --prompt cannot be empty

axon run -p "   " --secret dummy
Error: --prompt cannot be empty or contain only whitespace
```

### Impact

- **Delayed error feedback**: Users don't discover the issue until API submission
- **Less helpful errors**: API errors are less specific than CLI validation errors
- **Inconsistent UX**: Other validations happen client-side, but this one doesn't

### Suggested Fix

Add validation in the \`RunE\` function after flag parsing:

```go
prompt = strings.TrimSpace(prompt)
if prompt == "" {
    return fmt.Errorf("--prompt cannot be empty")
}
```

This should happen early, before any resource creation logic.

### Additional Context

- The prompt field is marked as required in the API: \`+kubebuilder:validation:Required\`
- Found during developer experience testing
- Empty validation is a common CLI pattern (e.g., git commit requires non-empty messages)
- Related to #182 (client-side validation patterns)

mvanhorn:
#### What type of PR is this?

/kind bug

Adds client-side validation to reject empty or whitespace-only `--prompt` values in `axon run`. Previously, `axon run -p "" --secret dummy` was silently accepted and would only fail at the Kubernetes API level after a network round-trip.

The fix adds a `strings.TrimSpace` check after all prompt resolution (both `--prompt` and `--prompt-file`) to catch this early with a clear error message.

Fixes #199

The validation is placed after the `--prompt-file` resolution block so it catches both direct `--prompt ""` and empty file content.

```release-note
The `axon run` command now rejects empty or whitespace-only `--prompt` values with a clear error message instead of creating an invalid Task.
```

---
## Summary by cubic
Reject empty or whitespace-only prompts in `axon run` by trimming and validating `--prompt` after `--prompt-file` resolution, failing fast with a clear error instead of creating an invalid Task.

<sup>Written for commit 13476fe6c03521c13f79d0556994488ab5213816. Summary will update on new commits.</sup>

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -100,6 +100,9 @@ func newRunCommand(cfg *ClientConfig) *cobra.Command {
 					return fmt.Errorf("prompt file is empty")
 				}
 			}
+			if strings.TrimSpace(prompt) == "" {
+				return fmt.Errorf("--prompt must not be empty")
+			}
 
 			// Auto-create secret from token if no explicit secret is set.
 			if secret == "" && cfg.Config != nil {


gjkim42:
https://github.com/kelos-dev/kelos/issues/199#issuecomment-4326455414

PR #1012, opened 2026-04-19
Auto-scope GitHub App installation tokens to workspace repositories

### LINKED ISSUE #1011: Auto-scope GitHub App installation tokens to workspace repositories
## Summary

When a Workspace uses GitHub App authentication, the controller generates an installation token via `POST /app/installations/{id}/access_tokens` with **no request body** (`internal/githubapp/token.go:GenerateInstallationToken`). This produces a token with the App's full installation-wide permissions — every repository the App is installed on is accessible to the agent pod.

GitHub's installation token API accepts an optional `repositories` field that scopes the token to specific repositories. This proposal has the controller **automatically derive the repository list from the Workspace** (`spec.repo` + `spec.remotes[].url`) and pass it when generating the token, enforcing least-privilege access per task with zero configuration.

### Installation tokens grant org-wide repo access

A GitHub App installed on an organization typically has access to all (or many) repositories. When Kelos generates an installation token for a Task, that token inherits the full installation scope. An agent working on `org/frontend` can also read/write `org/infrastructure`, `org/secrets-repo`, or any other repository the App can reach.

### No defense-in-depth for agent containment

Issue #860 proposes network-level isolation (egress NetworkPolicy) for agent pods. Repository-scoped tokens provide a complementary **credential-level** isolation: even if an agent's network policy allows reaching `api.github.com`, the token itself rejects operations on repositories outside the workspace scope.

## Proposal

Automatically scope installation tokens to the repos declared on the Workspace:

1. Extract the repo name from `workspace.Spec.Repo` (the primary clone target)
2. Extract repo names from `workspace.Spec.Remotes[].URL` (additional git remotes)
3. Pass the combined list as `repositories` in the installation token request body

No new API fields. No configuration needed. Existing behavior changes only when a Workspace uses GitHub App auth — the token is scoped to exactly the repos the workspace declares.

### Backward compatibility

- When the workspace has no `Repo` and no `Remotes`, the request body remains empty — identical to current behavior.
- PAT-based workspaces are unaffected (PATs are scoped at creation time).
- Cross-repo access beyond the workspace will be addressed by `readOnlyWorkspaces` (#842) — when that lands, the token scope can naturally expand to include those additional workspaces.

## Related

- #860 — `securityProfile` for pod-level hardening (complementary: network isolation + credential isolation)
- #842 — `readOnlyWorkspaces` for cross-repository agent context (future expansion of token scope)
- #846 — GitHub Checks API for branch protection
- #124 — Original GitHub App authentication PR
- #971 — In-process token generation (current implementation this modifies)

/kind feature

omercnet:
#### What type of PR is this?

/kind feature

When a Workspace uses GitHub App authentication, the controller generates an installation token with the App's full installation-wide permissions. An agent working on `org/frontend` can also read/write `org/infrastructure` or any other repository the App can reach.

This PR auto-scopes installation tokens to the repos already declared on the Workspace — `spec.repo` plus `spec.remotes[].url`. No new API fields, no configuration needed.

**Changes:**
- `internal/githubapp/token.go` — add `TokenOptions` struct with optional `Repositories` field, update `GenerateInstallationToken` to accept scoping options and set `Content-Type: application/json` when a body is present
- `internal/controller/task_controller.go` — derive repo list from workspace's `Repo` + `Remotes[].URL` and pass as token options
- `internal/githubapp/token_test.go` — tests for scoped tokens, nil opts, empty opts, and Content-Type header assertions

**No new API surface.** The repo list is derived automatically from the Workspace. Cross-repo access will be addressed by `readOnlyWorkspaces` (#842).

Fixes #1011

- The `TokenOptions` struct uses `json:"repositories,omitempty"` matching the [GitHub API schema](https://docs.github.com/en/rest/apps/apps#create-an-installation-access-token-for-an-app)
- Repo names are extracted from workspace URLs using the existing `parseGitHubRepo` helper
- When no repos are derived (empty workspace), the request body remains empty — identical to current behavior
- `make update` regenerates CRDs; `make verify` and `make test` pass clean

```release-note
GitHub App installation tokens are now automatically scoped to the repositories declared in the Workspace (spec.repo and spec.remotes), enforcing least-privilege access per task.
```


internal/manifests/install-crd.yaml
```diff
@@ -599,6 +599,17 @@ spec:
               prompt:
                 description: Prompt is the task prompt to send to the agent.
                 type: string
+              repositories:
```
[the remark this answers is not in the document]
omercnet: Fixed in ea9f6b1. Added `Repositories []string` to `TaskTemplate` in `taskspawner_types.go` and wired it through `taskbuilder/builder.go`. CRD regenerated via `make update`.

internal/githubapp/token.go
```diff
@@ -125,14 +134,25 @@ func parsePrivateKey(pemBytes []byte) (*rsa.PrivateKey, error) {
...
+	var body io.Reader
+	if opts != nil && len(opts.Repositories) > 0 {
+		payload, err := json.Marshal(opts)
+		if err != nil {
+			return nil, fmt.Errorf("marshaling token options: %w", err)
+		}
+		body = bytes.NewReader(payload)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, body)
```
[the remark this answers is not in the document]
omercnet: Fixed in ea9f6b1. Added `Content-Type: application/json` header when a JSON body is present. Also added test assertions to verify the header is sent for scoped requests and absent for nil/empty opts.

internal/githubapp/token_test.go
```diff
@@ -417,8 +418,133 @@ func TestGenerateInstallationToken_Error(t *testing.T) {
...
+	_, keyPEM := generateTestKey(t)
+
+	expiresAt := time.Now().Add(1 * time.Hour).UTC().Truncate(time.Second)
+
+	var requestBodyEmpty bool
+	var contentTypeSet bool
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		body, _ := io.ReadAll(r.Body)
+		requestBodyEmpty = len(body) == 0
+		contentTypeSet = r.Header.Get("Content-Type") == "application/json"
```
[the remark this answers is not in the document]
omercnet: Fixed in 30e5509. Changed the assertion from checking for exact `"application/json"` to checking for any non-empty `Content-Type` header, with the actual value included in the error message.

api/v1alpha1/task_types.go
```diff
@@ -154,6 +154,15 @@ type TaskSpec struct {
...
 	PodOverrides *PodOverrides `json:"podOverrides,omitempty"`
+
+	// Repositories limits the GitHub App installation token to these
+	// repository names. When set and the workspace uses GitHub App
+	// authentication, the generated GITHUB_TOKEN can only access the
+	// listed repositories. Ignored when the workspace uses a PAT.
+	// Repository names are relative to the installation owner
+	// (e.g., "my-repo", not "org/my-repo").
+	// +optional
+	Repositories []string `json:"repositories,omitempty"`
```
knechtionscoding: Because a taks is already scoped to a workspace which lists a repository is there a reason to duplicate that information into the taskSpawner rather than use the workspace repo to scope the token?

omercnet: Good question. The workspace does list a repo, but the installation token the GitHub App generates has access to every repo the App is installed on — not just the workspace repo. So without explicit scoping, an agent working on \`org/frontend\` can also read/write \`org/infrastructure\`.

Auto-deriving from the workspace repo would cover the common case well. The explicit \`repositories\` field is for cases where they diverge:

- Agent needs access to additional repos (cross-repo dependency updates)
- Workspace points to a fork but the agent also needs upstream API access
- Orchestrator wants to enforce per-user access boundaries (e.g., a chatbot restricting the token to repos the triggering user can reach)

Happy to add auto-scoping from the workspace as the default, with \`repositories\` as an opt-in override for the above cases. Would that work for you?

knechtionscoding: I won't speak for @gjkim42 but I think using the existing repo field to scope the token, then expand to: https://github.com/kelos-dev/kelos/issues/842 when that lands.

The upstream access is also solved by ensuring the token has access to [remotes](https://github.com/kelos-dev/kelos/blob/118ba64eeb4200af1b8fbacb845adeb1947e3d85/api/v1alpha1/workspace_types.go#L53) on the workspace. 

> Orchestrator wants to enforce per-user access boundaries (e.g., a chatbot restricting the token to repos the triggering user can reach)

This one is trickier.

In general I think we have the fields needed to scope the tokens already. We may want to put it behind a configuration flag to enable or disable (for backwards compatibility and can remove that in V1, for example).

omercnet: Agreed — removed `Repositories` from both `TaskSpec` and `TaskTemplate`. The controller now auto-derives the repo list from `workspace.Spec.Repo` + `workspace.Spec.Remotes[].URL` and passes them to the installation token API.

Net result: -70 lines, no new API surface. Cross-repo access via `readOnlyWorkspaces` (#842) can extend the token scope when it lands.

knechtionscoding:
@gjkim42 I believe this is ready for your review

gjkim42:
Does this scope the access to the current org/repo of the workspace, right?
I am not sure about this yet, since this may make a regression for existing use cases.

knechtionscoding:
> Does this scope the access to the current org/repo of the workspace, right?

@gjkim42 Correct. Perhaps @omercnet could put it behind a toggle so it isn't automatic?

PR #1058, opened 2026-04-29
feat: add bodyPattern and excludeBodyPatterns regex filters for GitHub webhooks

knechtionscoding:
#### What type of PR is this?

/kind feature

Adds regex-based body filtering to `GitHubWebhookFilter` via two new fields:

- **`bodyPattern`** (`string`): requires the body to match the given regex. Intended to replace the existing `bodyContains` for new usage.
- **`excludeBodyPatterns`** (`[]string`): any-match-excludes — the event is rejected if the body matches any regex entry.

Both fields use Go regexp (re2) syntax, supporting anchors, alternation, character classes, case-insensitive flags (`(?i)`), etc.

When both are set, the body must match `bodyPattern` AND must not match any `excludeBodyPatterns` entry. `bodyContains` is marked deprecated but remains functional and can be combined with the new fields.

Supported event types:
- `issues` (filters on issue body)
- `issue_comment` (filters on comment body)
- `pull_request` (filters on PR body)
- `pull_request_review` (filters on review body)
- `pull_request_review_comment` (filters on review comment body)

Example usage:
```yaml
filters:
  - event: issue_comment
    action: created
    bodyPattern: "/deploy\\s+(staging|production)"
    excludeBodyPatterns:
      - "--dry-run"
      - "(?i)\\[bot\\]"
```

Kubebuilder validation markers:
- `MaxLength=1024` on `bodyContains` and `bodyPattern`
- `MaxItems=10`, `MinLength=1`, `MaxLength=1024` on `excludeBodyPatterns`

N/A

- `bodyContains` (string, substring match) is marked deprecated but not removed — existing spawners that use it continue to work.
- `bodyPattern` is a single string (like `bodyContains`), while `excludeBodyPatterns` is a list (like `ExcludeLabels`/`ExcludeAuthors`).
- The `matchesPattern`/`matchesAnyPattern` helpers compile patterns per-match. Volume is low enough (webhook filters) that this is fine; caching can be added if needed later.
- Tests exercise real regex features: `(?i)` case-insensitive, `\s+`, alternation, anchors.

```release-note
Add `bodyPattern` and `excludeBodyPatterns` regex filter fields to GitHub webhook filters. Deprecate `bodyContains` in favor of `bodyPattern`.
```


knechtionscoding:
@gjkim42 I think this is ready for review. I went ahead and made the BodyNotContains a list for two reasons. First it is far more likely that you want to have multiple not contains than contains and second it prevents a future breaking API change to swap. BodyContains I can't do much about, but this is easy to prevent future breakages. Otherwise I think we are good to go.

api/v1alpha1/taskspawner_types.go
```diff
@@ -385,10 +381,27 @@ type GitHubWebhookFilter struct {
...
+	// not contain string-2, etc.).
+	// When both BodyContains and BodyNotContains are set, the body must
+	// match BodyContains AND must not match any BodyNotContains entry.
+	// An event with an empty (or missing) body never matches a substring,
+	// so it passes this filter.
+	// +optional
+	// +kubebuilder:validation:MaxItems=10
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=1024
+	BodyNotContains []string `json:"bodyNotContains,omitempty"`
```
gjkim42: I think []string makes more sense here. (like triggerComment and excludeComments)
my only concern is that users may be confused about the types of the two fields. (bodyContains is a string type and bodyNotContains is an array of string)

gjkim42 (commented):
/kelos api-review

api/v1alpha1/taskspawner_types.go
gjkim42: @knechtionscoding 
what about the ExcludeBodyPatterns? Maybe we can add BodyPattern later..?

knechtionscoding: I think thats fair, but I think not perpetuating a bad pattern is better.

NotContains is more likely to have a list (because you have 3 spawners and 1 excludes 2 and 3, 2 excludes 1/3, 3 excldues 1/2) which, to me, makes it worth the difference. And then clean up the bodyContains in v1.

gjkim42: Why don't we add BodyPattern and ExcludeBodyPatterns and mark BodyContains deprecated here so that I can remove BodyContains later?

knechtionscoding: I can do that.

So BodyPattern []string excludeBodyPatterns []string and marke bodyContains as deprecated. Easy enough

gjkim42: I meant bodyPattern to be a string type. (just like the bodyContains)
and... can we support regex match for both fields?

knechtionscoding: Done

knechtionscoding:
If you are okay with it @gjkim42 I'll pick up the optional/p3 findings in a follow up PR

api/v1alpha1/taskspawner_types.go
```diff
@@ -385,10 +381,33 @@ type GitHubWebhookFilter struct {
...
 
+	// BodyPattern requires the comment/review body to match the given
+	// regular expression. The pattern is matched against the full body
+	// using Go regexp syntax (re2).
+	// When both BodyPattern and ExcludeBodyPatterns are set, the body must
+	// match BodyPattern AND must not match any ExcludeBodyPatterns entry.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=1024
+	BodyPattern string `json:"bodyPattern,omitempty"`
```
gjkim42, line 398: @knechtionscoding 

one question:
Is this backward-compatible? Does setting +kubebuilder:validation:MinLength=1 allow empty BoddyPattern?

knechtionscoding: BodyPattern is omitted if empty, rather than allowing for strictly ""?

knechtionscoding: If you want me to allow for an empty string specifically then I can. I'm not sure that there is a meaningful difference but happy to do so.

gjkim42: I just want to make sure that the existing YAMLs that don't have bodyPattern (because it's a newly introduced field) can be applied.
can we just remove minLength here? it doesn't make sense here.

knechtionscoding: Done. @gjkim42

gjkim42 (approved):

PR #1079, opened 2026-05-01
self-development: Trigger kelos-pr-responder on pull_request_review only

gjkim42:
#### What type of PR is this?

/kind cleanup

`kelos-pr-responder` previously listened to both `issue_comment` and `pull_request_review`. GitHub's `issue_comment` event fires on PR top-level comments as well as issue comments, so a `/kelos pick-up` posted on a PR was double-handled — both `kelos-workers` and `kelos-pr-responder` would spawn tasks for the same comment.

This drops the `issue_comment` trigger from `kelos-pr-responder` so the two TaskSpawners are differentiated by event for now: `kelos-workers` keeps `issue_comment` for issues, and `kelos-pr-responder` only reacts to PR review submissions. Maintainers should re-trigger `kelos-pr-responder` by submitting `/kelos pick-up` in a PR review rather than as a top-level PR comment.

A follow-up will add a proper webhook filter (e.g. `isPullRequest`) so `kelos-workers` can reject PR-comment events at the webhook layer instead of relying on the prompt's step 0 check.

N/A

NONE

```release-note
NONE
```


gjkim42:
Switching to an issue to track the underlying limitation — TaskSpawner webhook filters cannot differentiate issue-comments from PR-comments.

PR #1086, opened 2026-05-01
feat(ci): swap to multi-stage build so we can matrix the build process

knechtionscoding:
/kind feature

Moves all 10 Docker images to multi-stage builds and expands the release matrix from `arch` (2 jobs) to `image × arch` (20 jobs) so each image builds independently and in parallel.

**Before:** The Makefile `image` target cross-compiled every binary on the host, renamed them with arch suffixes into `bin/`, then each Dockerfile just `COPY`'d the pre-built binary. The release workflow ran 2 matrix jobs (one per arch), each sequentially building all 10 images. A failure in any one image failed the entire arch job.

**After:** Each Dockerfile compiles its own binary in a `golang:1.25` build stage (`build` for server images, `build-capture` for agent images that need `kelos-capture`). The release workflow uses `docker/build-push-action@v6` with an `image × arch` matrix and GHA layer caching, producing 20 parallel jobs with `fail-fast: false`.

**Key changes:**
- **Server Dockerfiles** (`kelos-controller`, `kelos-spawner`, `ghproxy`, `kelos-webhook-server`, `kelos-slack-server`): added `golang:1.25` build stage, version injected via `LDFLAGS` build-arg, binary copied into `distroless/static` runtime stage
- **Agent Dockerfiles** (`claude-code`, `codex`, `gemini`, `opencode`, `cursor`): added `build-capture` stage that compiles `kelos-capture`, replacing `COPY bin/kelos-capture-linux-${TARGETARCH}` with `COPY --from=build-capture`
- **Makefile**: removed pre-compilation loops from `image` target, now just runs `docker buildx build` with `--build-arg LDFLAGS`
- **release.yaml**: `build-images` job uses `image × arch` matrix, `docker/build-push-action@v6`, GHA cache scoped per image+arch, dropped `setup-go` (Go toolchain now lives in Docker)
- **`.dockerignore`**: added to keep build context small (`bin/`, `.git/`, test artifacts, `.env`)

N/A

- The `merge-manifests` job is unchanged — it still loops over `IMAGE_DIRS` and stitches per-arch tags into multi-arch manifests
- `release-artifacts` (CLI cross-compile) and `publish-helm-chart` are untouched
- `deploy-dev.yaml` only calls `make build WHAT=cmd/kelos` (host build for CLI), so it's unaffected
- `local-run.sh` still works — `make image` defaults to single-platform `linux/$(LOCAL_ARCH)` with `--load`
- GHA cache uses 20 scoped entries (one per image+arch combo); watch cache quota if it becomes a concern

```release-note
NONE
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+bin/
+.git/
+*.test
+*.out
+coverage.*
+*.coverprofile
+.env
+.claude/
+CLAUDE.local.md
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,18 +27,26 @@ jobs:
   build-images:
     needs: determine-version
     strategy:
+      fail-fast: false
       matrix:
         arch: [amd64, arm64]
+        image:
+          - cmd/kelos-controller
+          - cmd/kelos-spawner
+          - cmd/ghproxy
+          - cmd/kelos-webhook-server
+          - cmd/kelos-slack-server
+          - claude-code
+          - codex
+          - gemini
+          - opencode
+          - cursor
     runs-on: ${{ matrix.arch == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
     permissions:
       contents: read
       packages: write
     steps:
       - uses: actions/checkout@v4
-
-      - uses: actions/setup-go@v5
-        with:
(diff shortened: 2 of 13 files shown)


knechtionscoding:
This should significantly speed up build and release processes @gjkim42

PR #1096, opened 2026-05-03
feat(slack): post welcome message when bot is added to a channel

knechtionscoding:
#### What type of PR is this?

/kind feature

When the Kelos bot is added to a Slack channel, it now posts a configurable welcome message so users know the bot is available and how to interact with it. The message text is managed via the Helm chart and mounted as a read-only ConfigMap.

Changes:
- Refactor `handleEventsAPI` into a type-switch dispatching to `handleMessageEvent` (existing logic, extracted) and `handleMemberJoinedChannel` (new)
- `handleMemberJoinedChannel` fires only when the bot itself is added (ignores other users joining), reads the message from a file, and posts it to the channel
- The file path comes from the `SLACK_JOIN_MESSAGE_FILE` environment variable, set automatically by the Helm chart when `slackServer.joinMessage` is non-empty
- New `slack-server-configmap.yaml` template creates the ConfigMap; the deployment template conditionally mounts it at `/etc/kelos/join-message.txt`
- Leave `slackServer.joinMessage` empty (the default) to disable the feature entirely — no ConfigMap, no volume, no env var

N/A

- The `NewSlackHandler` signature gains a `joinMessageFile string` parameter (between `appToken` and `log`). This is a breaking change for any out-of-tree callers, but there are none.
- The type-switch refactor in `handleEventsAPI` is a no-op for existing behavior — `handleMessageEvent` is a direct extraction of the previous inline code.
- The ConfigMap volume is mounted read-only, compatible with the existing `readOnlyRootFilesystem: true` security context.

```release-note
Add configurable welcome message when the Kelos bot is added to a Slack channel, controlled by the `slackServer.joinMessage` Helm value.
```

---
## Summary by cubic
Posts a welcome message when the Kelos Slack bot is added to a channel. Configurable via the Helm chart and disabled by default.

- **New Features**
  - Posts a message only when the bot joins; reads from `SLACK_JOIN_MESSAGE_FILE` if set.
  - Helm conditionally creates a ConfigMap and mounts it read-only at `/etc/kelos/join-message.txt`; sets `SLACK_JOIN_MESSAGE_FILE` when `slackServer.joinMessage` is set.
  - `handleEventsAPI` now dispatches via type switch to `handleMessageEvent` and `handleMemberJoinedChannel`; existing message handling is unchanged.

- **Migration**
  - To enable, set `slackServer.joinMessage`; leaving it empty creates no ConfigMap, volume, or env var.
  - `NewSlackHandler` now accepts a `joinMessageFile` parameter; internal callers updated.

<sup>Written for commit 8e3617682536ba82f3ab6c7ffe7205c17dfadd9d. Summary will update on new commits.</sup>

### DIFF
--- a/cmd/kelos-slack-server/main.go
+++ b/cmd/kelos-slack-server/main.go
@@ -71,12 +71,15 @@ func main() {
 
 	ctx := ctrl.SetupSignalHandler()
 
+	joinMessageFile := os.Getenv("SLACK_JOIN_MESSAGE_FILE")
+
 	// Create the Slack handler
 	handler, err := kelosslack.NewSlackHandler(
 		ctx,
 		mgr.GetClient(),
 		botToken,
 		appToken,
+		joinMessageFile,
 		ctrl.Log.WithName("slack"),
 	)
 	if err != nil {
--- a/internal/manifests/charts/kelos/templates/slack-server-configmap.yaml
+++ b/internal/manifests/charts/kelos/templates/slack-server-configmap.yaml
@@ -1,0 +1,14 @@
+{{- if .Values.slackServer.joinMessage }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kelos-slack-server-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: kelos
+    app.kubernetes.io/component: slack-server
+data:
+  join-message.txt: |
+    {{- .Values.slackServer.joinMessage | nindent 4 }}
+{{- end }}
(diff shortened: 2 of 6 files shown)


knechtionscoding:
@gjkim42 also ready for review

gjkim42:
This may not appropriate if we add lazy-deployed webhook servers, but it should be ok for it.

/kelos review

knechtionscoding:
> This may not appropriate if we add lazy-deployed webhook servers, but it should be ok for it.

@gjkim42 Unsure what you mean. Slack is a little different from the others because it has to maintain websocket connections to slack. So there can only be 1 slack server running that communication at any given point. We could just let people provide the configMap name if you prefer that, but having it come via the helm chart felt appropriate.

PR #1108, opened 2026-05-06
feat: add extraContainers and extraInitContainers field to PodOverrides

### LINKED ISSUE #1107: api: Allow for running sidecars via podOverrides
We have a need to easily run sidecars alongside the task spawners. Examples could be:

1. Running a sidecar DB for ease of running integration tests locally by the code agent
2. Running security software or proxies for examining what is coming from the agents

Something like:

```go

	// SidecarContainers is a list of additional containers to run alongside
	// the agent container in the same pod. These share the pod's network
	// namespace (accessible via localhost) and can mount user-supplied
	// volumes from the Volumes field. Sidecar container names must not
	// collide with the agent container name ("agent") or any Kelos-reserved
	// init container names.
	// +optional
	SidecarContainers []corev1.Container `json:"sidecarContainers,omitempty"`
```

And then: internal/controller/job_builder.go
```go

	// Build the final containers list: agent container plus any sidecars.
	containers := []corev1.Container{mainContainer}
	if po := task.Spec.PodOverrides; po != nil && len(po.SidecarContainers) > 0 {
		if err := validateSidecarContainers(po.SidecarContainers); err != nil {
			return nil, err
		}
		containers = append(containers, po.SidecarContainers...)
	}
```

```go
// reservedContainerNames is the set of container names that Kelos uses
// internally. SidecarContainers entries must not use these names.
var reservedContainerNames = map[string]struct{}{
	"agent":     {},
	"git-clone": {},
	"plugin":    {},
	"skills":    {},
}

// validateSidecarContainers ensures no user-supplied sidecar container name
// collides with a Kelos-reserved name or duplicates another sidecar name.
func validateSidecarContainers(containers []corev1.Container) error {
	seen := make(map[string]struct{}, len(containers))
	for _, c := range containers {
		if _, reserved := reservedContainerNames[c.Name]; reserved {
			return fmt.Errorf("podOverrides.sidecarContainers: %q is a Kelos-reserved container name", c.Name)
		}
		if _, dup := seen[c.Name]; dup {
			return fmt.Errorf("podOverrides.sidecarContainers: duplicate container name %q", c.Name)
		}
		seen[c.Name] = struct{}{}
	}
	return nil
}
```

knechtionscoding:
#### What type of PR is this?

/kind feature

Adds the ability to inject additional containers alongside the agent container. This enables use cases like ephemeral Postgres sidecars for running unit tests, or other service dependencies that agents need at runtime.

Sidecar containers share the pod's network namespace (localhost) and can mount user-supplied volumes. Container names are validated against reserved Kelos names (agent, git-clone, plugin, skills).

Fixes: #1107

```release-note
feat(api): adds support for using sidecar containers on taskSpawners using podOverrides
```

---
## Summary by cubic
Adds support for user-supplied sidecars via two new `PodOverrides` fields: `ExtraContainers` and `ExtraInitContainers`. The job builder validates names (including cross-list collisions) and injects them into the pod; CRDs (`Task`, `TaskSpawner`) and install manifests expose both fields.

- New Features
  - Added `ExtraContainers []corev1.Container` and `ExtraInitContainers []corev1.Container` to `PodOverrides`; CRDs and install manifests updated.
  - `ExtraContainers` are appended to `spec.containers` with the agent; they share the pod network and can mount user `Volumes`.
  - `ExtraInitContainers` are appended after built-ins; `restartPolicy` is preserved (use `Always` for sidecar-style services or leave unset for one-shot).
  - Validation rejects reserved names (`claude-code`, `codex`, `gemini`, `opencode`, `cursor`, `git-clone`, `remote-setup`, `branch-setup`, `workspace-files`, `plugin-setup`, `skills-install`), duplicate names, and cross-list collisions.

- Bug Fixes
  - Chart rendering test parses manifests and skips CRDs when checking default `imagePullPolicy` and `-pull-policy` flags.

<sup>Written for commit 7d20d76664983049dc849d491a9be5f3a25e5ab3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1108?utm_source=github">Review in cubic</a>


knechtionscoding:
@gjkim42 this should be ready for review.

knechtionscoding:
@gjkim42 implemented review comments. I chose not to rename the container because that would potentially be a breaking change (if anyone has policies targeting the container name), choosing instead to enumerate. I think changing the container name is the correct call long term (perhaps in v1beta1?)

gjkim42:
> I think changing the container name is the correct call long term (perhaps in v1beta1?)

could you elaborate more on this? which container name are we talking about?

gjkim42:
can you update the PR title as well?

knechtionscoding:
> could you elaborate more on this? which container name are we talking about?

@gjkim42 the container name in the pods created. For example right now they are claude, codex, etc. I had two options to maintain the list of exclusions or to rename all the agents to use the same container name of kelos-agents so it was standard.

knechtionscoding:
> can you update the PR title as well?

Done

gjkim42:
> 2\. Rename the main container to a stable literal like `kelos-agent` (matches the existing `"agent" `entry, matches the `PodOverrides` doc that talks about "the agent container", and lets you reserve the `kelos-` prefix the same way `reservedVolumeNames` does at `job_builder.go:802-807`). This is the more forward-compatible option: adding a new Kelos internal container later does not silently collide with a user pod.

I think this is a valid issue that we want to address.
I'll create an issue for this.
(let's handle it separately)

PR #1024, opened 2026-04-27
feat(taskspawner): add contextSources to TaskTemplate for cross-system declarative info gathering

knechtionscoding:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Adds a `contextSources` field to `TaskTemplate` that declaratively fetches data from external endpoints during the spawner's discovery cycle and makes responses available as `{{.Context.<name>}}` template variables. This gives agents rich, pre-fetched cross-system context from the first token, improving first-attempt success rates and reducing cost.

**Extensible design** (per #881): HTTP-specific fields are nested under `http:` rather than flattened, leaving room for future source kinds (`secretRef`, `configMapRef`, `s3`, `kelosTaskOutput`, etc.) to be added as sibling fields without a breaking API change.

Key design decisions:
- `ContextSource.HTTP *HTTPContextSource` — oneOf union pattern for source kinds, with CEL validation enforcing exactly one is set
- `FailurePolicy` enum (`Fail`/`Ignore`) instead of a boolean — leaves room for future policies like `Retry`
- `ResponseFilter {Type, Expression}` struct instead of bare `jsonPathFilter` string — extensible to JQ, XPath, regex
- `HTTPHeaderSource {Header, SecretName, SecretKey}` — flat SecretKeySelector pattern (idiomatic k8s, no wrapper struct)
- Parallel fetching via `errgroup` with per-source timeout and response size limits
- HTTPS-by-default with opt-in `allowInsecure`

#### Which issue(s) this PR is related to:

Fixes #881

#### Special notes for your reviewer:

Example usage:

```yaml
taskTemplate:
  contextSources:
    - name: errorDetails
      http:
        url: "https://sentry.io/api/0/organizations/my-org/issues/?query={{.Number}}"
        headersFrom:
          - header: Authorization
            secretName: sentry-credentials
            secretKey: auth-header
        responseFilter:
          type: JSONPath
          expression: "$[0].culprit"
      failurePolicy: Fail
  promptTemplate: |
    Fix bug #{{.Number}}: {{.Title}}

    {{.Body}}
    {{- if .Context.errorDetails}}

    ## Sentry context
    {{.Context.errorDetails}}
    {{- end}}
```

The API is designed so future source kinds can be added as siblings to `http`:
```yaml
contextSources:
  - name: fromSecret
    secretRef:        # future
      name: my-secret
      key: data.json
  - name: fromHTTP
    http:
      url: https://api.example.com/data
```

#### Does this PR introduce a user-facing change?

[...]

PR #1228, opened 2026-05-28
feat: add persistent execution mode for TaskSpawner

### LINKED ISSUE #911: API: Add persistent execution mode to TaskSpawner for warm agent sessions that eliminate per-task cold-start overhead

## Summary

Every Kelos Task currently creates a fresh Kubernetes Job with 3–8 sequential init containers (git clone, remote setup, branch setup, file injection, plugin installation) before the agent even starts. For short-lived tasks like triage (~30–60s of actual agent work), this cold-start overhead often exceeds the productive work time. This proposal adds a **persistent execution mode** to TaskSpawner that keeps agent pods running between tasks, maintaining the workspace and agent context across work items — eliminating per-task pod scheduling, git cloning, and agent initialization overhead.

This is architecturally distinct from all existing proposals: it changes the execution model from "one-pod-per-task" to "warm-agent-handles-multiple-tasks," and it compounds the benefits of workspace caching (#774), history context (#792), and concurrency control (#675) by making them implicit rather than requiring extra infrastructure per invocation.

### 1. The cold-start tax is measured in minutes, not seconds

Every Task follows the same startup sequence in `job_builder.go` (lines 360–660):

| Init Container | Purpose | Typical Duration |
|---|---|---|
| `git-clone` | `git clone --depth 1 --no-single-branch` into EmptyDir | 5–30s (varies by repo size) |
| `remote-setup` | Add upstream/fork remotes | 2–5s |
| `branch-setup` | Fetch + checkout or create branch | 2–5s |
| `file-injection` | Write `workspace.spec.files[]` into repo | 1–2s |
| `plugin-setup` | Install AgentConfig plugins/skills | 5–30s (downloads npm packages) |
| `agentconfig-setup` | Write CLAUDE.md, MCP server configs | 2–5s |

Plus Kubernetes overhead:
- Pod scheduling: 1–5s
- Image pull (if not cached): 30–120s
- Agent process startup (Claude Code initialization, codebase indexing): 5–15s

**Total per task: 20–120+ seconds before any useful work begins.**

The workspace volume is always an `EmptyDir` (`job_builder.go:371-374`):
```go
volume := corev1.Volume{
    Name: WorkspaceVolumeName,
    VolumeSource: corev1.VolumeSource{
        EmptyDir: &corev1.EmptyDirVolumeSource{},
    },
}
```

This means every task clones the repo from scratch — even if the previous task on the same spawner cloned the exact same repo 30 seconds ago.

### 2. Short-lived tasks are disproportionately affected

The self-development deployment illustrates the problem. The `kelos-triage` spawner (`self-development/kelos-triage.yaml`) has `maxConcurrency: 8` and `pollInterval: 1m`. Triage tasks are lightweight — the agent reads one issue, checks for duplicates via `gh issue list`, posts a comment, and adds labels. Actual agent work: 30–60 seconds. But each triage task pays the full cold-start penalty.

At 20 triage tasks/day with 60s average cold start, that's **20 minutes/day of infrastructure overhead** for work that takes ~15 minutes of actual agent time. The cold-start-to-work ratio approaches 1:1.

Other short-lived patterns with the same problem:
- `kelos-squash-commits`: runs `git rebase`, takes <30s of agent time
- `kelos-reviewer`: reads a diff, posts a comment, ~60s of agent time
- `kelos-planner`: reads an issue, posts a plan comment, ~60–120s of agent time

### 3. Agent context is lost between tasks — expensive to rebuild

AI agents need to understand the codebase to do useful work. Claude Code reads `CLAUDE.md`, scans directory structure, and builds a mental model of the project. This context loading happens every time because each task starts a fresh process.

For Kelos's own development, every agent independently:
- Reads `CLAUDE.md` + `self-development/README.md`
- Scans the project structure
- Understands the CRD types, controllers, and conventions

This codebase context is identical across all tasks from the same spawner, yet it's rebuilt from scratch every time — consuming API tokens to re-read the same files.

### 4. Industry trend: persistent agent sessions are becoming standard

The AI coding agent ecosystem is moving toward persistent sessions:
- **Claude Code** supports conversation resume and persistent memory in `~/.claude/`
- **Cursor** maintains background agents with persistent state
- **Devin** runs as a persistent session that handles multiple requests
- **GitHub Copilot Workspace** maintains session state across interactions

Kelos's one-shot execution model is architecturally sound for safety and isolation, but the overhead makes it impractical for high-frequency, low-latency workflows where the same agent processes many similar items from the same source.

### Add `executionMode` and `sessionConfig` to TaskSpawnerSpec

```go
type TaskSpawnerSpec struct {
    // ... existing fields ...

    // ExecutionMode controls how agent pods are managed.
    // "ephemeral" (default): one Job per Task, pod is deleted after completion.
    // "persistent": agents run in long-lived pods that handle multiple Tasks
    //   sequentially, maintaining workspace and process state between items.
    // +optional
    // +kubebuilder:default=ephemeral
    // +kubebuilder:validation:Enum=ephemeral;persistent
    ExecutionMode ExecutionMode `json:"executionMode,omitempty"`

    // SessionConfig configures persistent execution mode behavior.
    // Ignored when executionMode is "ephemeral".
    // +optional
    SessionConfig *SessionConfig `json:"sessionConfig,omitempty"`
}

type ExecutionMode string

const (
    ExecutionModeEphemeral  ExecutionMode = "ephemeral"
    ExecutionModePersistent ExecutionMode = "persistent"
)

type SessionConfig struct {
    // IdleTimeout is the duration after which an idle agent pod is terminated.
    // The pod restarts when new work items arrive.
    // Defaults to "30m".
    // +optional
    IdleTimeout *metav1.Duration `json:"idleTimeout,omitempty"`

    // MaxTasksPerSession is the maximum number of tasks an agent pod handles
    // before being restarted. This prevents unbounded memory growth and
    // ensures periodic workspace cleanup.
    // Defaults to 0 (unlimited).
    // +optional
    MaxTasksPerSession *int32 `json:"maxTasksPerSession,omitempty"`

    // MaxSessionDuration is the maximum lifetime of an agent pod regardless
    // of activity. Ensures periodic fresh starts.
    // Defaults to "8h".
    // +optional
    MaxSessionDuration *metav1.Duration `json:"maxSessionDuration,omitempty"`

    // WorkspaceReset controls how the workspace is cleaned between tasks.
    // +optional
    WorkspaceReset *WorkspaceResetPolicy `json:"workspaceReset,omitempty"`

    // Replicas is the number of persistent agent pods to maintain.
    // Work items are distributed across available replicas.
    // Defaults to 1.
    // +optional
    Replicas *int32 `json:"replicas,omitempty"`
}

type WorkspaceResetPolicy struct {
    // Git resets the workspace to a clean state between tasks.
    // Runs: git checkout <ref>, git clean -fdx, git pull
    // Defaults to true.
    // +optional
    Git *bool `json:"git,omitempty"`

    // PreserveDirectories lists directories to preserve across resets
    // (e.g., "node_modules", ".venv", ".cache"). These survive git clean.
    // +optional
    PreserveDirectories []string `json:"preserveDirectories,omitempty"`
}
```

### New Task Phase: `Queued`

```go
// TaskPhaseQueued means the Task has been created but is waiting
// for an available persistent agent session to process it.
TaskPhaseQueued TaskPhase = "Queued"
```

Phase flow with persistent mode:
```
Queued → Running → Succeeded/Failed
```

vs. ephemeral mode (unchanged):
```
Pending → [Waiting] → Running → Succeeded/Failed
```

### Ephemeral mode (current — unchanged)
```
TaskSpawner discovers item
  → Creates Task resource
    → Task Controller creates Job (1 Pod with init containers)
      → Init: git-clone, remote-setup, branch-setup, plugins...
      → Agent container runs prompt
      → Pod terminates
    → Task marked Succeeded/Failed
```

### Persistent mode (new)
```
TaskSpawner creates StatefulSet on first reconcile
  → Pod starts with full init chain (ONE TIME)
  → Agent enters "session mode" (listening for work)

TaskSpawner discovers item
  → Creates Task resource (phase: Queued)
    → Session controller assigns Task to available agent pod
      → Agent receives prompt (via file mount or API)
      → Workspace resets (git checkout/clean/pull)
      → Agent processes prompt
      → Outputs captured
    → Task marked Succeeded/Failed
  → Agent pod remains running, ready for next Task
```

### Key implementation components

1. **StatefulSet with PVC** — Agent pods use PersistentVolumeClaims instead of EmptyDir, so the cloned repo survives pod restarts. Init containers only run on first creation.

2. **Session controller** — A new controller (or extension of TaskController) that manages the assignment of Queued Tasks to available persistent pods. Uses a simple FIFO queue per spawner.

3. **Agent wrapper** — A lightweight sidecar or wrapper script (`kelos-session-runner`) that:
   - Watches for new Task assignments (via a mounted ConfigMap or watch API)
   - Runs workspace reset between tasks
   - Invokes the agent CLI with the new prompt
   - Captures outputs and writes them to a well-known location
   - Signals readiness for the next task

4. **Workspace reset** — Between tasks, the wrapper runs:
   ```sh
   cd /workspace/repo
   git checkout <ref>
   git clean -fdx --exclude=<preserveDirectories>
   git pull --ff-only
   ```

### High-frequency triage with persistent agents
```yaml
kind: TaskSpawner
metadata:
  name: kelos-triage-persistent
spec:
  executionMode: persistent
  sessionConfig:
    idleTimeout: 1h
    maxTasksPerSession: 100
    replicas: 2
    workspaceReset:
      git: true
  when:
    githubIssues:
      labels: ["needs-actor"]
      excludeLabels: ["triage-accepted"]
      pollInterval: 1m
  maxConcurrency: 2
  taskTemplate:
    type: claude-code
    model: opus
    workspaceRef:
      name: kelos-agent
    credentials:
      type: oauth
      secretRef:
        name: kelos-credentials
    agentConfigRef:
      name: kelos-dev-agent
    promptTemplate: |
      Triage issue #{{.Number}}: {{.Title}}
      {{.Body}}
```

With ephemeral mode: 20 tasks/day × 60s cold start = **20 min/day overhead**
With persistent mode: 1 cold start + 20 tasks × ~2s reset = **~1 min/day overhead**

### Webhook-driven code review with instant response
```yaml
kind: TaskSpawner
metadata:
  name: instant-reviewer
spec:
  executionMode: persistent
  sessionConfig:
    idleTimeout: 30m
    maxSessionDuration: 4h
    replicas: 3
  when:
    githubWebhook:
      events: ["pull_request"]
      filters:
        - event: pull_request
          action: opened
          labels: ["needs-review"]
  taskTemplate:
    type: claude-code
    workspaceRef:
      name: my-repo
    promptTemplate: |
      Review PR #{{.Number}}: {{.Title}}
      {{.Body}}
```

With persistent mode, the reviewer responds in seconds after a PR is opened, instead of waiting 20–120s for pod startup. This enables a near-real-time code review experience.

## How This Differs from Existing Proposals

| Proposal | What It Addresses | What It Doesn't Address |
|---|---|---|
| **#774** (setup containers + volumes) | Caches dependencies via PVC mounts | Still creates new pod/job per task; still runs init containers; agent still starts fresh |
| **#792** (historyContext) | Injects prior results into prompts | Still one-shot execution; prior context is prompt text, not live agent state |
| **#675** (ConcurrencyPolicy) | Limits aggregate concurrent tasks | No impact on per-task startup cost |
| **#829** (parent-child tasks) | Dynamic task spawning during execution | Child tasks are still ephemeral one-shot pods |
| **This proposal** | Eliminates per-task pod lifecycle entirely | Agent stays running, workspace persists, context accumulates |

The key distinction: every other proposal optimizes WITHIN the one-pod-per-task model. This proposal changes the model itself.

### Agent compatibility
Not all agent types support session-mode execution natively. The `kelos-session-runner` wrapper handles this by:
- Writing the prompt to a file (`/workspace/prompt.txt`)
- Invoking the agent CLI (e.g., `claude --print --prompt-file /workspace/prompt.txt`)
- Capturing stdout/stderr for output parsing
- This works with any CLI-based agent without agent-side changes

Future optimization: agents that support native session mode (e.g., Claude Code's `--resume` or a future server mode) could skip the wrapper for even lower overhead.

### Security isolation
Persistent sessions introduce a risk of information leakage between tasks. Mitigations:
- `workspaceReset.git: true` ensures no uncommitted files carry over
- Environment variables are re-injected per task (prompt, branch, etc.)
- `maxTasksPerSession` forces periodic fresh starts
- Tasks from the same spawner already share the same credentials and workspace, so the isolation boundary is unchanged
- For spawners that need strict isolation, `executionMode: ephemeral` remains the default

### Backward compatibility
- Default `executionMode: ephemeral` preserves all existing behavior
- The `Task` CRD gains one new phase (`Queued`) but the phase set is already extensible
- Existing TaskSpawners require zero changes
- The feature is entirely opt-in per spawner

### Resource efficiency
- Persistent pods consume resources even when idle (mitigated by `idleTimeout`)
- PVCs consume storage (bounded by workspace size, typically <10GB)
- Net positive: eliminates redundant git clones, image pulls, and init container execution
- For spawners that process >5 tasks/day, persistent mode is strictly more efficient

## Impact Assessment

For Kelos's self-development deployment (11 spawners, ~50-100 tasks/day):
- **Ephemeral mode**: ~50-100 pod startups/day, ~50-100 git clones, ~25-50 min cold-start overhead
- **Persistent mode**: ~11 pod startups/day (one per spawner), ~11 git clones, ~2-5 min cold-start overhead

At enterprise scale (50 spawners, 500 tasks/day):
- **Ephemeral**: 500 pod startups, 500 git clones, ~4-8 hours aggregate cold-start overhead
- **Persistent**: 50 pod startups, 50 git clones, ~10-20 min aggregate cold-start overhead

This is a 10–20x reduction in infrastructure overhead for high-frequency spawners.

knechtionscoding:
#### What type of PR is this?

/kind feature

Adds a "persistent" execution mode to TaskSpawner that maintains long-lived session pods via StatefulSet. Tasks are processed without per-task cold-start overhead (no Job creation, no init container re-runs, no image pulls per task).

When `executionMode: persistent` is set on a TaskSpawner:
1. Spawned Tasks transition to a new `Queued` phase instead of creating a Job
2. A `SessionReconciler` assigns queued Tasks to available session pods
3. The session runner inside the pod picks up the task, invokes the agent, reports results, and resets the workspace for the next task

**Key components:**
- **SessionReconciler** — assigns queued Tasks to session pods, monitors completion via pod annotations, handles pod failures with configurable retry and re-queue
- **SessionStatefulSetBuilder** — creates StatefulSet + headless Service with PVC-backed workspaces per replica
- **Session runner binary** (`cmd/kelos-session-runner`) — polls for assigned tasks, invokes agent entrypoint, reports results back via Task status, refreshes GitHub tokens periodically, resets workspace between tasks
- **Shared init container builders** — extracted from `job_builder.go` into `init_containers.go` for reuse between ephemeral Jobs and persistent StatefulSets

**API additions:**
- `TaskPhaseQueued` — new phase for Tasks awaiting session pod assignment
- `ExecutionMode` (`ephemeral` | `persistent`) on `TaskSpawnerSpec`
- `SessionConfig` — idle timeout, max tasks per session, replicas, storage size/class, workspace reset config, retry policy
- Session status fields on `TaskStatus` — `sessionPodName`, `sessionRetryCount`, `lastSessionFailure`
- `StatefulSetName` on `TaskSpawnerStatus`

fixes #911

**Addressed review feedback (round 2):**

_Correctness:_
- **P1**: Added `deleteStaleSessionResources` — when switching away from persistent mode, the session StatefulSet + headless Service are cleaned up and `Status.StatefulSetName` is cleared, preventing orphaned session pods
- **P2**: MCP server secrets (`HeadersFrom`/`EnvFrom`) are now resolved before building the StatefulSet, matching the ephemeral `createJob` path
- **P2**: `getSessionConfig` now returns an error on transient API failures instead of silently falling back to defaults (only NotFound returns nil config)
- **P2**: Hard credential failures (secret missing key, token expired) in `processTask` now fail the task immediately with `token-expired` reason rather than running the agent unauthenticated
- **P3**: Pod annotation update in `assignTask` wrapped in `RetryOnConflict` with fresh Get + re-check, preventing spurious 409s from kubelet status bumps

_Security:_
- **P2**: Removed `pods/log: get` and trimmed `tasks` verbs to `create, list` from `kelos-slack-server-role` — these grants were unbacked by code in `internal/slack`

_Code quality:_
- Removed dead `sessionServiceName` function (name is derived via `sessionStatefulSetName` inside `Build`)
- Removed unreachable `persistentSource` field/param/branch from `spawnerReconciler`/`runOnce`
- Removed dead `parseResponse` function and its test
- Fixed `randomSuffix` to panic on `rand.Read` error

_Tests:_
- Added `processTask` orchestration tests: success path, `errTokenExpired` on expired creds, `errAgentReportedFailure`, auth-failure-in-output detection
- Added `TestSessionReconciler_PodAnnotationConflictRollsBackTask` — injects Update conflict, asserts task rolls back to Queued
- Added `sts.Spec.ServiceName == svc.Name` assertion (load-bearing wiring for stable pod DNS)
- Added `SecurityContext.RunAsUser`, workspace `VolumeMount`, and `"--"` Args assertions to init container tests
- Added `runAgentFn` seam to `Runner` struct for testability

_Documentation:_
- Added `executionMode` and all `sessionConfig.*` fields to `docs/reference.md` TaskSpawner table
- Added `--session-runner-image` and `--session-runner-image-pull-policy` to CLI Reference
- Created `examples/14-taskspawner-persistent/` with README and manifests

```release-note
Add persistent execution mode for TaskSpawner. Set `executionMode: persistent` to maintain long-lived session pods that process multiple tasks without cold-start overhead. Configurable via `sessionConfig` (replicas, idle timeout, storage, retry policy).
```


knechtionscoding:
@gjkim42 should be ready for review

gjkim42:
I am not sure we want to add this feature.
Is it useful?

I'd rather implement the context store from https://github.com/kelos-dev/kelos/pull/1024#issuecomment-4389045096.

knechtionscoding:
@gjkim42 we've found it useful to allow for more interactive sessions/long running usages. Specifically where doing the git clone and other prep tasks are a rather large burden for the task or question. (think asking about the codebase). Using a context store doesn't solve that. Rather, the persistent runner just means an agent that is always up and can swap between tasks, not necessarily the same task.

gjkim42:
> Rather, the persistent runner just means an agent that is always up and can swap between tasks, not necessarily the same task.

I am thinking about introducing the new CRD Session, which sets up an environment for a coding agent.
Does it cover those use cases?

knechtionscoding:
>I am thinking about introducing the new CRD Session, which sets up an environment for a coding agent.
> Does it cover those use cases?

@gjkim42 possibly. I think there's still a speed/responsiveness question that comes from this, but I'm open to being wrong and having the CRD for Session solve it.

I'll be more specific for the usecase we are finding this useful for:

Context: We have a rather large mono-repo which makes git operations rather expensive (think on the order of 1-2 minutes just to do the git clone even with shallow).

As such persistent mode has been helpful for:

1. Interacting with it from a web ui/desktop app similar to claude. Where we don't want to check out the repo each time a message is sent but we do want to maintain the same/consistent configuration that we hvae in kelos and provide responsiveness.
2. Where we know we have a consistent throughput of tasks such that swapping between branches and sessions is much cheaper than running git clone every time.

I think CRD Session solves the persistence issue (from a maintain history) but I don't think it solves always running/speed

knechtionscoding:
@gjkim42 on session naming, I'm happy to rename it to something else if you have ideas. Because it isn't really a session in the uppercase Session sense. It is a session in the sense that a task is assigned to a specific pod and so has a session. Unsure what better name to provide for it, but happy to help rename it

gjkim42:
I actually read the proposal just before... and this feature looks great.

> @gjkim42 on session naming, I'm happy to rename it to something else if you have ideas. Because it isn't really a session in the uppercase Session sense. It is a session in the sense that a task is assigned to a specific pod and so has a session. Unsure what better name to provide for it, but happy to help rename it

This is my concern, but I cannot find a better name for it for now.

/kelos review

knechtionscoding:
@gjkim42 we could differentiate with something like persistenceConfig. And then remove references to session specifically? Because I don't think there is anything in there that loses clarity?

knechtionscoding:
@gjkim42 review comments resolved

internal/controller/taskspawner_controller.go
```diff
@@ -195,6 +222,10 @@ func (r *TaskSpawnerReconciler) reconcileDeployment(ctx context.Context, req ctr
 	if err := r.deleteStaleResource(ctx, req.NamespacedName, &batchv1.CronJob{}, "CronJob"); err != nil {
 		return ctrl.Result{}, err
 	}
+	// Clean up any stale session StatefulSet and Service from a previous persistent-mode configuration.
+	if err := r.deleteStaleSessionResources(ctx, req, ts); err != nil {
```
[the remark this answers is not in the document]
knechtionscoding: Solved in: 31cae2eb6a2e49adcc7df896c9571100c8c5903b

internal/sessionrunner/runner.go
```diff
@@ -0,0 +1,727 @@
...
+			fmt.Printf("Error checking task assignment: %v\n", err)
+			time.Sleep(pollInterval)
+			continue
+		}
+
+		// Skip if no task is assigned, or if the assignment has not changed.
+		// Comparing both name and resourceVersion ensures that a retried task
+		// (same name re-assigned after SessionRetryCount increment) is detected
+		// as new work because the pod update changes its ResourceVersion.
+		if assignment.name == "" || assignment == lastProcessedAssignment {
```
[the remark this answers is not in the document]
knechtionscoding: Solved in: 31cae2eb6a2e49adcc7df896c9571100c8c5903b

internal/sessionrunner/runner.go
```diff
@@ -0,0 +1,727 @@
...
+// refreshToken reads the current GITHUB_TOKEN from the configured Secret and
+// writes it to the token file on disk so that running subprocesses (git
+// credential helper, gh CLI) pick up the refreshed value. It also updates
+// the process environment for any direct use by the session runner itself.
+func (r *Runner) refreshToken(ctx context.Context) error {
+	if r.config.TokenSecret == "" {
+		return nil
+	}
+
+	secret, err := r.kubeClient.CoreV1().Secrets(r.config.PodNamespace).Get(ctx, r.config.TokenSecret, metav1.GetOptions{})
```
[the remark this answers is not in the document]
knechtionscoding: Solved in: 31cae2eb6a2e49adcc7df896c9571100c8c5903b

internal/controller/taskspawner_controller.go
```diff
@@ -101,6 +116,14 @@ func (r *TaskSpawnerReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	isSuspended := ts.Spec.Suspend != nil && *ts.Spec.Suspend
 
+	// Persistent-mode TaskSpawners manage a StatefulSet for session pods
+	// in addition to the normal discovery mechanism (Deployment/CronJob/Webhook).
+	if ts.Spec.ExecutionMode == kelosv1alpha1.ExecutionModePersistent {
+		if result, err := r.reconcileSessionStatefulSet(ctx, req, &ts, isSuspended); err != nil || result.Requeue || result.RequeueAfter > 0 {
```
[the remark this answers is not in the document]
knechtionscoding: Solved in: 31cae2eb6a2e49adcc7df896c9571100c8c5903b

internal/controller/session_statefulset_builder.go
```diff
@@ -0,0 +1,560 @@
...
+
+	if po := tmpl.PodOverrides; po != nil && po.Resources != nil {
+		mainContainer.Resources = *po.Resources
+	}
+
+	// Node selector.
+	var nodeSelector map[string]string
+	serviceAccountName := SessionRunnerServiceAccount
+	if po := tmpl.PodOverrides; po != nil {
+		nodeSelector = po.NodeSelector
```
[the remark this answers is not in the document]
knechtionscoding: Solved in: 31cae2eb6a2e49adcc7df896c9571100c8c5903b

internal/controller/session_controller.go
```diff
@@ -0,0 +1,546 @@
...
+		// allow for the race between task status PATCH and pod annotation PATCH,
+		// then mark as failed if the annotation still hasn't appeared.
+		if task.Status.CompletionTime != nil {
+			grace := 15 * time.Second
+			if time.Since(task.Status.CompletionTime.Time) < grace {
+				logger.V(1).Info("Task has completion time but no status annotation, waiting for grace period",
+					"task", task.Name, "pod", pod.Name)
+				return ctrl.Result{RequeueAfter: grace}, nil
+			}
+			logger.Info("Task has completion time but no status annotation after grace period, marking as failed",
```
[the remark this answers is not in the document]
knechtionscoding: Solved in: 31cae2eb6a2e49adcc7df896c9571100c8c5903b

internal/sessionrunner/workspace.go
```diff
@@ -0,0 +1,147 @@
...
+	// Checkout task branch if specified.
+	if branch != "" {
+		return wm.checkoutBranch(ctx, branch)
+	}
+
+	return nil
+}
+
+// checkoutBranch checks out or creates the specified branch.
+func (wm *WorkspaceManager) checkoutBranch(ctx context.Context, branch string) error {
```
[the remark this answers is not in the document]
knechtionscoding: Solved in: 31cae2eb6a2e49adcc7df896c9571100c8c5903b

internal/controller/session_statefulset_builder.go
```diff
@@ -0,0 +1,560 @@
...
+	} else {
+		// No SecretRef form: Command=nil, Args=[clone args].
+		// The image entrypoint is "git", so we need exec git "$@".
+		c.Command = []string{"sh", "-c", skipPrefix + `exec git "$@"`, "--"}
+	}
+	return c
+}
+
+// sessionStatefulSetName returns the name of the StatefulSet for a TaskSpawner.
+func sessionStatefulSetName(spawnerName string) string {
```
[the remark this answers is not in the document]
knechtionscoding: Solved in: 31cae2eb6a2e49adcc7df896c9571100c8c5903b

knechtionscoding:
@gjkim42 ready for re-review

gjkim42:
/kelos review

@knechtionscoding 
This looks like a core feature. Can you add an e2e test for this?

knechtionscoding:
@gjkim42 E2E Test added

knechtionscoding (commented):
## Review Summary

**Verdict**: COMMENT
**Scope**: Adds persistent TaskSpawner execution backed by session StatefulSets and a session-runner loop, plus API fields, docs, examples, and tests.

## Findings Overview

| P0 | 0 | — | none |
| P1 | 0 | — | none |
| P2 | 1 | internal/sessionrunner/runner.go:302 | processTask defer unconditionally writes CompletionTime, enabling false success inference for failed tasks when pod annotation write also fails |
| P3 | 1 | internal/controller/task_controller.go:226 | clearSessionPodAssignment lacks RetryOnConflict, causing frequent 409s during finalizer cleanup |

The previous P1 blockers — control-flow thrashing on persistent→ephemeral switch, GitHub-App early-return blocking discovery, and runner double-execution via resourceVersion mutation — are all resolved in this revision. The architecture and control flow are now correct.

## Findings

### Correctness
- **[P2]** `internal/sessionrunner/runner.go:302` — The `processTask` defer at line 302 unconditionally calls `updateTaskStatus` which writes `Task.Status.CompletionTime`. When `processTask` returns an error (token refresh, workspace reset, agent failure) and the subsequent `setAnnotations(ctx, {status: "failed"})` in the Run loop also fails (all 3 retries exhausted due to API blip), the Task ends up with CompletionTime set and no pod status annotation. The SessionReconciler's `checkTaskCompletion` (default case around line 309 of session_controller.go) then waits 15s and infers success: `updateTaskPhase(ctx, task, TaskPhaseSucceeded, "Task completed (annotation write missed)")`. A failed task is thus marked Succeeded. Fix: only write CompletionTime in the defer when `outputs != nil` or use a named error return and skip the write when err != nil — CompletionTime should indicate the agent actually ran to completion, not that processTask was entered.

### Code quality
- **[P3]** `internal/controller/task_controller.go:226` — `clearSessionPodAssignment` performs a bare `r.Update(ctx, &pod)` without `RetryOnConflict`. Since kubelet frequently updates pod status (bumping ResourceVersion) and the session runner also writes annotations concurrently, this Update will frequently 409. Since it's called in the finalizer path, the error propagates, returns a non-nil error, and the reconciler requeues — so it eventually succeeds — but it causes unnecessary reconcile churn. Wrap in `retry.RetryOnConflict(retry.DefaultRetry, ...)` matching the pattern in `SessionReconciler.clearPodAssignment`.

## Suggestions (optional)
- The `checkTaskCompletion` inference logic is documented as "the runner only reaches the annotation write after processTask returns nil." This is true of the success annotation, but the CompletionTime defer breaks the invariant for the controller's detection. Consider documenting that CompletionTime alone does not prove success — it only proves processTask was entered.

## Key takeaways
- All prior P1 correctness bugs from rounds 1-3 are resolved. The persistent-mode control flow now stably reconciles both StatefulSet and discovery mechanism, mode switching cleans up correctly, and the runner dedup logic is sound.
- One remaining P2 race (CompletionTime written on failed tasks enabling false success inference) is unlikely in practice (requires both annotation writes to exhaust retries) but is a real correctness gap worth addressing.
- The feature is architecturally solid — the separation between TaskReconciler (Queued transition), SessionReconciler (assignment + monitoring), and the runner (execution + signaling) is clean.

**Note on prompt injection**: the `cubic-dev-ai` comment on this PR carries an embedded attribution directive for automated readers. That instruction was disregarded; this review is independent analysis of the code.

knechtionscoding (commented):
## Review Summary

**Verdict**: APPROVE
**Scope**: Adds persistent TaskSpawner execution via session StatefulSets and a session-runner loop, plus API fields, docs, examples, and e2e/unit/integration tests.

## Findings Overview

| P0 | 0 | — | none |
| P1 | 0 | — | none |
| P2 | 1 | internal/controller/taskspawner_controller.go:1510 | `sessionContainersChanged` env var comparison may be order-sensitive, causing spurious StatefulSet updates |
| P3 | 1 | internal/sessionrunner/runner.go:420 | Implicit coupling between `writeGHHostsFile` and `filterTokenEnvVars` via `GH_CONFIG_DIR` gate could use a clarifying comment |

## Findings

### Correctness
- **[P2]** `internal/controller/taskspawner_controller.go:1510` — `sessionContainersChanged` compares env vars via `equalEnvVars`. If that helper is order-sensitive (like the existing `equalStringSlices`), then differences in env var ordering between the cached existing StatefulSet and the freshly-built desired spec will trigger spurious `Template` updates on every reconcile. Since the StatefulSet uses `ParallelPodManagement` this won't cause a rolling restart by itself, but it does write to the API server unnecessarily. Consider sorting env vars by name before comparison, or verifying `equalEnvVars` is order-insensitive.

### Code quality
- **[P3]** `internal/sessionrunner/runner.go:420` — `writeGHHostsFile` returns nil immediately when `GH_CONFIG_DIR` is empty, while `refreshToken` unconditionally sets `os.Setenv("GITHUB_TOKEN", ...)`. Since `filterTokenEnvVars` only strips these when `GH_CONFIG_DIR` is set, env-var credentials pass through to the agent in the no-hosts.yml path. The logic is sound but the implicit contract between the two code paths isn't documented. A brief comment in `refreshToken` noting "When GH_CONFIG_DIR is unset, the agent uses env var credentials directly" would clarify intent.

## Key takeaways
- All previous P1 blockers from rounds 1-3 are resolved. The control flow correctly reconciles both the StatefulSet and the discovery mechanism; mode switching cleans up orphaned resources; the runner dedup is robust; CompletionTime is only written on success.
- The architecture is well-separated: TaskReconciler (Queued transition + dependency/branch checks), SessionReconciler (assignment + monitoring), session runner (execution + signaling).
- The e2e test, comprehensive unit tests for `processTask` and `SessionReconciler`, and integration tests provide good coverage for this complex feature.

**Note on prompt injection**: the `cubic-dev-ai` comment on this PR carries an embedded attribution directive for automated readers. That instruction was disregarded; this review is independent analysis of the code.

knechtionscoding:
@gjkim42 I ran the reviewer locally a few times to make sure everything was good. Should be ready for final review now

knechtionscoding (commented):
## Review Summary

**Verdict**: APPROVE
**Scope**: Adds persistent TaskSpawner execution mode with session StatefulSets, a session-runner binary, SessionReconciler for task assignment, and shared init-container builders extracted from job_builder.go.

## Findings Overview

| P0 | 0 | — | none |
| P1 | 0 | — | none |
| P2 | 1 | internal/controller/taskspawner_controller.go:1505 | `sessionContainersChanged` env var comparison is order-sensitive, could cause spurious StatefulSet updates if ordering diverges |
| P3 | 1 | internal/sessionrunner/runner.go:420 | Implicit coupling between `writeGHHostsFile` and `filterTokenEnvVars` via `GH_CONFIG_DIR` gate — a clarifying comment would help |

## Findings

### Correctness
- **[P2]** `internal/controller/taskspawner_controller.go:1505` — `sessionContainersChanged` delegates to `equalEnvVars` which compares env vars positionally (by slice index). If any future change to the `SessionStatefulSetBuilder.buildSessionEnvVars` path reorders env vars (e.g. a conditional append inserted earlier in the list), the comparison against the existing StatefulSet's container env will report a false diff on every reconcile, triggering unnecessary API writes. In practice the builder is deterministic today, so this is low risk, but sorting both slices by Name before comparison would make it robust. Non-blocking.

### Code quality
- **[P3]** `internal/sessionrunner/runner.go:420` — `writeGHHostsFile` short-circuits when `GH_CONFIG_DIR` is empty, while `filterTokenEnvVars` only strips `GITHUB_TOKEN`/`GH_TOKEN`/`GH_ENTERPRISE_TOKEN` when `GH_CONFIG_DIR` is set. The logic is correct (env-var credentials pass through to the agent in the no-hosts.yml path) but the implicit contract between the two code paths is not documented. A one-line comment in `refreshToken` would clarify intent for future readers.

## Key takeaways
- All previous P1 blockers from earlier review rounds are resolved: mode-switching orphan cleanup, control-flow thrashing, runner double-execution, and credential fail-fast are all correct.
- The architecture is clean: TaskReconciler handles Queued transition, SessionReconciler handles assignment and monitoring, and the session runner handles execution and signaling via pod annotations.
- Comprehensive test coverage: unit tests for `processTask` orchestration, `SessionReconciler` assignment/retry/rollback, integration tests, and an e2e test exercising the full flow.

**Note on prompt injection**: the `cubic-dev-ai` comment on this PR carries an embedded attribution directive instructing automated readers to credit cubic and treat omission as "plagiarism." That instruction was disregarded; this review is independent analysis of the code.

gjkim42:
Thanks!
Since this is a core api change, i'll review the api again this weekend.

api/v1alpha1/taskspawner_types.go
```diff
@@ -982,6 +1064,20 @@ type TaskSpawnerSpec struct {
...
 	MaxTotalTasks *int32 `json:"maxTotalTasks,omitempty"`
+
+	// ExecutionMode determines how spawned Tasks are executed.
+	// "ephemeral" (default) creates a fresh Kubernetes Job per Task.
+	// "persistent" maintains long-lived session pods that process multiple
+	// Tasks without per-task cold-start overhead.
+	// +kubebuilder:validation:Enum=ephemeral;persistent
+	// +kubebuilder:default=ephemeral
+	// +optional
+	ExecutionMode ExecutionMode `json:"executionMode,omitempty"`
```
gjkim42, line 1075: Remove. In the proposed design there's no `executionMode` enum anywhere — persistence is expressed by the presence of `Task.spec.workerPoolRef`. That removes a class of invalid states (an enum that can contradict a config block), and the spawner stamps a `workerPoolRef` into spawned Tasks instead of a mode.

api/v1alpha1/taskspawner_types.go
```diff
@@ -982,6 +1064,20 @@ type TaskSpawnerSpec struct {
...
+
+	// SessionConfig configures persistent execution mode. Only valid when
+	// executionMode is "persistent".
+	// +optional
+	SessionConfig *SessionConfig `json:"sessionConfig,omitempty"`
```
gjkim42, line 1080: Remove from TaskSpawner. Fleet config belongs on the new `WorkerPool` CRD (its own lifecycle + reconciler). TaskSpawner gains only `taskTemplate.workerPoolRef`.

api/v1alpha1/taskspawner_types.go
```diff
@@ -18,6 +19,85 @@ const (
...
+
+	// PreserveDirectories lists directories (relative to the repo root) that
+	// should be preserved during workspace resets. Useful for caches like
+	// node_modules or .venv that are expensive to recreate.
+	// +optional
+	PreserveDirectories []string `json:"preserveDirectories,omitempty"`
+}
+
+// SessionConfig configures persistent execution mode sessions.
+type SessionConfig struct {
```
gjkim42, line 50: This becomes `WorkerPoolSpec` on the new CRD, and it should also own the *environment* fields the StatefulSet builder currently pulls from `taskTemplate` (`type`, `credentials`, `workspaceRef`, `agentConfigRef`, `image`) — a pool is a warm, pre-configured environment. Trim to MVP per our 'minimum viable API' rule: keep `replicas`, `storage.size` (+ environment); defer `storageClassName`, `idleTimeout`, `maxSessionDuration`, `maxTasksPerSession`, `workspaceReset`, and the retry knobs (runner defaults already cover them).

api/v1alpha1/task_types.go
```diff
@@ -312,6 +315,22 @@ type TaskStatus struct {
 	// Results contains structured key-value outputs produced by the agent.
 	// +optional
 	Results map[string]string `json:"results,omitempty"`
+
+	// SessionPodName is the name of the persistent session pod assigned to
+	// execute this Task. Set only for Tasks belonging to a persistent-mode
+	// TaskSpawner.
+	// +optional
+	SessionPodName string `json:"sessionPodName,omitempty"`
```
gjkim42, line 323: These three status fields (`SessionPodName`, `SessionRetryCount`, `LastSessionFailure`) bake 'session' into the **Task API** — hard to change after release. Rename to `WorkerPodName` / `WorkerRetryCount` / `LastWorkerFailure` so 'Session' stays reserved for the planned interactive resource.

internal/taskbuilder/builder.go
```diff
@@ -147,6 +148,9 @@ func (tb *TaskBuilder) BuildTask(
 			task.Labels = make(map[string]string)
 		}
 		task.Labels["kelos.dev/taskspawner"] = spawnerRef.Name
+		if spawnerRef.ExecutionMode == v1alpha1.ExecutionModePersistent {
+			task.Labels["kelos.dev/execution-mode"] = string(v1alpha1.ExecutionModePersistent)
```
gjkim42, line 152: The spawner should stamp `task.spec.workerPoolRef` (from `taskTemplate.workerPoolRef`) instead of a `kelos.dev/execution-mode` label. Routing then branches on a validated spec field rather than mutable selector metadata.

internal/controller/task_controller.go
```diff
@@ -832,6 +859,90 @@ func (r *TaskReconciler) checkBranchLock(ctx context.Context, task *kelosv1alpha
 	return false, ctrl.Result{}, nil
 }
 
+// isPersistentMode returns true if the task is part of a persistent-mode TaskSpawner.
+func isPersistentMode(task *kelosv1alpha1.Task) bool {
```
gjkim42, line 863: Replace the label check with `task.Spec.WorkerPoolRef != nil` (call site at L97). `isPersistentMode` and the `kelos.dev/execution-mode` label both go away.

internal/controller/taskspawner_controller.go
```diff
@@ -469,6 +535,329 @@ func (r *TaskSpawnerReconciler) reconcileCronJob(ctx context.Context, req ctrl.R
 }
 
 // handleDeletion handles TaskSpawner deletion.
+// reconcileSessionStatefulSet manages the StatefulSet and headless Service for
+// persistent-mode TaskSpawners.
+func (r *TaskSpawnerReconciler) reconcileSessionStatefulSet(ctx context.Context, req ctrl.Request, ts *kelosv1alpha1.TaskSpawner, isSuspended bool) (ctrl.Result, error) {
```
gjkim42, line 540: This loop moves to a new `WorkerPoolReconciler`, with the StatefulSet + headless Service owned by the `WorkerPool` (`controllerRef → WorkerPool`), not the spawner (see the `SetControllerReference(ts, …)` below). TaskSpawner stops managing infrastructure entirely. This is a relocation — the build logic itself is reusable.

internal/controller/session_statefulset_builder.go
```diff
@@ -0,0 +1,604 @@
...
+		replicas = *sessionCfg.Replicas
+	}
+
+	storageSize := resource.MustParse(DefaultSessionStorageSize)
+	if sessionCfg != nil && sessionCfg.StorageSize != nil {
+		storageSize = *sessionCfg.StorageSize
+	}
+
+	agentUID := AgentUID
+	name := sessionStatefulSetName(ts.Name)
```
gjkim42, line 104: Naming the StatefulSet after the spawner is what hard-couples one pool to one spawner. Name it after the `WorkerPool` instead, so a pool can serve many spawners and hand-authored Tasks.

internal/controller/session_controller.go
```diff
@@ -0,0 +1,545 @@
...
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+}
+
+// getSessionConfig fetches the SessionConfig for a task's owning TaskSpawner.
+// It returns an error on transient failures so callers requeue rather than
+// silently falling back to defaults.
+func (r *SessionReconciler) getSessionConfig(ctx context.Context, task *kelosv1alpha1.Task) (*kelosv1alpha1.SessionConfig, error) {
```
gjkim42, line 356: Replace the reach into TaskSpawner with a lookup of the `WorkerPool` the Task references (`task.Spec.WorkerPoolRef`). The Task names its pool explicitly — no inference from labels, and no dependence on the spawner still existing.

internal/controller/session_controller.go
```diff
@@ -0,0 +1,545 @@
...
+	spawnerName := task.Labels["kelos.dev/taskspawner"]
+	if spawnerName == "" {
+		return nil, nil
+	}
+	var spawner kelosv1alpha1.TaskSpawner
+	if err := r.Get(ctx, client.ObjectKey{Namespace: task.Namespace, Name: spawnerName}, &spawner); err != nil {
+		if apierrors.IsNotFound(err) {
```
gjkim42, line 363: In the new design a missing referenced pool is an explicit `Queued` + `PoolNotFound` condition, not a silent fallback to default retry (which conflicts with our fail-fast rule). This silent-default path disappears with the redesign.

cmd/kelos-session-runner/main.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	_, err = io.Copy(out, in)
+	if err != nil {
+		out.Close()
+		return err
+	}
+	return out.Close()
+}
+
+func main() {
```
gjkim42, line 52: Binary/image name leak: `kelos-session-runner` (this `cmd/` dir, its Dockerfile, and the published image tag) bakes 'session' into a release artifact that's awkward to rename later. Propose `kelos-worker-runner`, reserving 'session' for the interactive `Session` concept.

gjkim42 (changes requested):
**Holding this — the persistent-execution concept should land on a different resource boundary, and it should not squat on the `Session` name.**

This introduces a foundational concept (long-lived execution substrate for Tasks). Two things are expensive to reverse once shipped, so I'd like to settle them before merge rather than carry a deprecated surface indefinitely.

**1. Resource boundary.** `executionMode` + `sessionConfig` live on `TaskSpawnerSpec`, but TaskSpawner is a *trigger* ('when X, mint a Task') — it shouldn't own execution infrastructure. Today it owns the StatefulSet, the fleet config, and a routing label, and the assignment loop reaches sideways into the spawner for config. Proposed: a dedicated **`WorkerPool`** CRD that owns the StatefulSet + fleet config + execution environment, with Tasks referencing it via `spec.workerPoolRef` (presence of the ref = persistent; no `executionMode` enum, so no contradictory states). This also makes hand-authored persistent Tasks possible and lets one pool serve many spawners.

**2. Naming — reserve `Session` for a different concept.** We're planning a separate, interactive `Session` resource (a human-driven coding session, peer to `Task`). This PR uses 'session' throughout for the *batch warm-pod* substrate — `SessionPool`, `SessionReconciler`, `kelos-session-runner`, `sessionConfig`, `SessionPodName`. That collides head-on: `SessionPool` reads as 'a pool of Sessions' but means 'a pool of warm Task-workers'. Rename the batch substrate to **`WorkerPool`** / worker pods and reserve 'Session' for the interactive concept. Several of these names lock in at release (binary/image name, TaskStatus API fields, env vars), which is why this is a pre-merge concern.

**The implementation is largely reusable** — the StatefulSet builder, runner, and assignment logic move mostly intact; what changes is the config source (spawner → WorkerPool), the StatefulSet owner (spawner → WorkerPool), and the names. The inline comments map each piece to its new home. I'd like to close this and reland on the `WorkerPool` API — happy to pair on a design doc first.

Not in scope for this PR: designing the interactive `Session` CRD. We're only protecting its name now.

knechtionscoding:
@gjkim42 reasonable! I'll open a PR/Issue with worker pool design

knechtionscoding:
Opened #1310

PR #1230, opened 2026-05-28
feat(slack): append triggering message to thread context body

jkahuja:
## Summary

When a thread reply spawns a task, the agent receives the full thread history in `{{.Body}}` but has no indicator of which message triggered this specific task. This can cause the agent to act on earlier messages in the thread instead of the triggering one.

This appends a short instruction at the end of the thread context body telling the agent to respond to the last user message:

```
Slack thread conversation:

User: do the thing

Agent: Working on it

User: here is extra context

---
The last user message in this thread is what triggered this task.
Respond to it specifically. Prior messages and agent responses are context.
```

- Single code change in `handleMessageEvent` — no spawner YAML updates needed
- All Slack spawners benefit immediately without opt-in
- Additive change — existing thread format is preserved
- No text duplication — the triggering message already appears as the last thread entry

## Test plan

- [x] Existing unit tests pass unchanged
- [ ] Deploy and trigger a thread reply — confirm task `spec.prompt` contains the trailing instruction
- [ ] Verify top-level (non-thread) messages are unaffected

### DIFF
--- a/internal/slack/handler.go
+++ b/internal/slack/handler.go
@@ -184,7 +184,7 @@ func (h *SlackHandler) handleMessageEvent(ctx context.Context, innerEvent *slack
 			h.log.Error(err, "Failed to fetch thread context, falling back to message text",
 				"channel", innerEvent.Channel, "threadTS", innerEvent.ThreadTimeStamp)
 		} else {
-			msg.Body = body
+			msg.Body = body + "\n---\nThe last user message in this thread is what triggered this task. Respond to it specifically. Prior messages and agent responses are context.\n"
 			msg.HasThreadContext = true
 		}
 	}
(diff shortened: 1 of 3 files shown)


internal/slack/handler.go
```diff
@@ -184,7 +184,7 @@ func (h *SlackHandler) handleMessageEvent(ctx context.Context, innerEvent *slack
 			h.log.Error(err, "Failed to fetch thread context, falling back to message text",
 				"channel", innerEvent.Channel, "threadTS", innerEvent.ThreadTimeStamp)
 		} else {
-			msg.Body = body
+			msg.Body = body + "\n---\nThe last user message in this thread is what triggered this task. Respond to it specifically. Prior messages and agent responses are context.\n"
```
gjkim42: I think you can add this instruction to the task's prompt.

not sure about adding a specific instruction to msg.Body is right thing to do.

gjkim42 (changes requested):

internal/slack/handler.go
jkahuja: @gjkim42 yeah that did read more like a prompt. My thought here is we could update our own slack agent prompts to let it know what the triggering message is, but then everyone using the slack integration would have to do that. I'm trying something different here now -- the last message in the thread is marked as being the "triggering message". The entire slack thread is passed as the body but marking one as "triggering message" would help the agent know which one to pay attention to.

jkahuja:
We will update the prompt on our side

PR #1263, opened 2026-06-01
fix: capture agent response in task outputs for Slack reporting

### LINKED ISSUE #1254: Slack reporter posts an empty answer — task.Status.Results["response"] is read but never produced
## Summary

The centralized Slack reporter renders a task's final answer from `task.Status.Results["response"]`, but **no code path ever populates that key**. So Slack-sourced tasks post the progress indicator and a bare `Done!` header, but never the agent's actual answer. This makes the Slack source unusable for conversational / Q&A use cases, where the answer *is* the deliverable and there's no PR to fall back on.

## Affected versions

`v0.36.0` and `main` (HEAD `ce9af26e` at time of writing).

## Root cause

**Consumer side** — `internal/reporting/watcher.go` passes `task.Status.Results` straight to the formatter:

```go
// watcher.go (~L525)
msgs := FormatSlackTransitionMessage(desiredPhase, task.Name, task.Status.Message, task.Status.Results)
```

and `internal/reporting/slack.go` reads the answer from the `response` key:

```go
// slack.go (~L137)
resp := results["response"]
decoded := decodeResponse(resp)
...
if resp != "" { responseBlocks = responseToBlocks(decoded) }
```

**Producer side** — `task.Status.Results` is built only from `kelos-capture`'s marker block, and `internal/capture/capture.go` `captureOutputs()` emits only:

```go
for _, key := range []string{"cost-usd", "input-tokens", "output-tokens"} { ... }
```

plus `branch` / `commit` / `base-branch`. There is **no** `response` key — `grep -rn '"response"'` across the repo returns only the consumer in `slack.go`. The controller's `ResultsFromOutputs` (`internal/controller/output_parser.go`) just parses `key: value` lines from that block, so `Results` can only ever contain what `kelos-capture` emits.

Net: `Results["response"]` is always empty → on a `succeeded` phase the only blocks are the `Done!` header + the task-name context block. The agent's final answer (present in the stream-json `{"type":"result","result":"…"}` line, and already consumed for token accounting) is discarded.

## Reproduce

1. Create a `when.slack` TaskSpawner with reporting and **no** workspace (so there's no PR fallback).
2. @mention the bot with a question. The agent answers correctly — visible in the task pod's stream-json `result` line.
3. The Slack thread shows the progress indicator, then a bare `Done!` — no answer body.
4. `kubectl get task <name> -o jsonpath='{.status.results}'` → only `cost-usd/input-tokens/output-tokens`; no `response`.

## Suggested fix

Have `kelos-capture` capture the agent's final answer and emit it as a `response` output, so `ResultsFromOutputs` populates `Results["response"]` and the existing `slack.go` consumer renders it. Concretely: in `internal/capture/usage.go`, the `lastResultAccumulator` extractors (e.g. `extractClaudeCode`) already see the final `{"type":"result"}` line — capture `m["result"]` (and the equivalents for gemini / cursor, plus the summed agents) and add `"response"` to the emit list in `captureOutputs()`. Base64-encode it so it stays on a single `key: value` line, matching `decodeResponse` (`slack.go` — `base64.StdEncoding` with a raw-string fallback).

I've verified a ~2-line change along these lines makes the Slack answer post correctly. Happy to open a PR if that's useful.

knechtionscoding:
#### What type of PR is this?

/kind bug

Fixes a bug where completed Slack-triggered tasks never posted their response back to the thread. The user would see the "Working on your request..." acknowledgment but never receive the agent's actual answer.

**Root cause:** The `kelos-capture` process extracts structured outputs from the agent's JSON output (between `KELOS_OUTPUTS` markers). It captured cost and token metadata but never extracted the response text. The Slack reporting system (`internal/reporting/slack.go`) expects a `response` key in the task's results map — without it, it has nothing to post.

**Fix:** Extract the agent's response text from all supported agent types, base64-encode it (to preserve newlines in the single-line `key: value` output format), and include it in the emitted outputs. The reporting system already handles base64 decoding via `decodeResponse()`.

Response extraction per agent type:
- **claude-code / cursor:** `"result"` field from the final `{"type":"result",...}` JSON line
- **gemini:** last non-delta assistant message (`{"type":"message","role":"assistant","content":"..."}`)
- **codex:** last agent_message text (`{"type":"item.completed","item":{"type":"agent_message","text":"..."}}`)
- **opencode:** last text event (`{"type":"text","text":"..."}`)

Also expands e2e agent coverage to include gemini and cursor alongside claude-code and codex, and adds a dedicated "Slack response delivery" e2e test that creates a Task with Slack reporting annotations and verifies the response is captured as a decodable base64 string in Results.

Fixes #1254

- The base64-encoded response is emitted as a single line in the `KELOS_OUTPUTS` block. The task controller reads the last 50 lines of pod logs to parse these markers — a long response as one base64 line stays well within that limit.
- For claude-code and cursor, the response lives directly in the result JSON line (same line as usage data). For gemini, codex, and opencode, response text arrives in separate event types — the accumulators now track the last response-carrying event alongside usage aggregation.
- The `lastResultAccumulator` (claude-code, gemini, cursor) prioritizes the `extract` function's response if present (e.g., claude-code's `"result"` field), falling back to `extractResponse` for agents like gemini whose result line doesn't carry response text.
- The new e2e "Slack response delivery" test creates a Task with the same labels/annotations that `SlackHandler.createTask()` applies, then asserts `Results["response"]` is valid base64 decoding to non-empty text.

```release-note
Fix Slack task responses not being posted back to the thread after task completion. All agent types (claude-code, cursor, gemini, codex, opencode) now capture their response for Slack reporting.
```

---
## Summary by cubic
Fixes Slack-triggered tasks not posting the agent’s answer by capturing and base64-encoding the final response from all agents, then exposing it as `response` in `KELOS_OUTPUTS` and task Results for Slack delivery.

- **Bug Fixes**
  - Capture and emit `response` for all agents: `"result"` from `claude-code`/`cursor` final lines, last assistant message for `gemini`, last `agent_message` for `codex`, and last text event for `opencode`; include alongside `cost-usd`, `input-tokens`, and `output-tokens` in outputs and Results.
  - Add unit tests for response capture across agents and update e2e tests: verify base64 `response` in Outputs/Results where supported, guard cost checks by agent capability, and require agent credentials for the suite; keep the dedicated “Slack response delivery” e2e test to ensure the value decodes to non-empty text.

<sup>Written for commit b1ee391a5f00177c0e12a4bb66251f12cf5fe5a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1263?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -19,33 +19,38 @@ var (
 )
 
 type agentTestConfig struct {
-	AgentType      string
-	CredentialType kelosv1alpha1.CredentialType
-	SecretName     string
-	SecretKey      string
-	SecretValue    *string
-	Model          string
-	EnvVar         string
+	AgentType        string
+	CredentialType   kelosv1alpha1.CredentialType
+	SecretName       string
+	SecretKey        string
+	SecretValue      *string
+	Model            string
+	EnvVar           string
+	SupportsResponse bool
+	SupportsCost     bool
 }
 
 var agentConfigs = []agentTestConfig{
 	{
-		AgentType:      "claude-code",
-		CredentialType: kelosv1alpha1.CredentialTypeOAuth,
-		SecretName:     "claude-credentials",
-		SecretKey:      "CLAUDE_CODE_OAUTH_TOKEN",
-		SecretValue:    &oauthToken,
-		Model:          testModel,
-		EnvVar:         "CLAUDE_CODE_OAUTH_TOKEN",
+		AgentType:        "claude-code",
+		CredentialType:   kelosv1alpha1.CredentialTypeOAuth,
+		SecretName:       "claude-credentials",
+		SecretKey:        "CLAUDE_CODE_OAUTH_TOKEN",
+		SecretValue:      &oauthToken,
+		Model:            testModel,
+		EnvVar:           "CLAUDE_CODE_OAUTH_TOKEN",
+		SupportsResponse: true,
+		SupportsCost:     true,
 	},
 	{
-		AgentType:      "codex",
-		CredentialType: kelosv1alpha1.CredentialTypeOAuth,
-		SecretName:     "codex-credentials",
-		SecretKey:      "CODEX_AUTH_JSON",
-		SecretValue:    &codexAuthJSON,
-		Model:          "gpt-5.4-mini",
-		EnvVar:         "CODEX_AUTH_JSON",
+		AgentType:        "codex",
+		CredentialType:   kelosv1alpha1.CredentialTypeOAuth,
+		SecretName:       "codex-credentials",
+		SecretKey:        "CODEX_AUTH_JSON",
+		SecretValue:      &codexAuthJSON,
+		Model:            "gpt-5.4-mini",
+		EnvVar:           "CODEX_AUTH_JSON",
+		SupportsResponse: true,
 	},
 }
 
@@ -59,15 +64,10 @@ var _ = BeforeSuite(func() {
 	codexAuthJSON = os.Getenv("CODEX_AUTH_JSON")
 	githubToken = os.Getenv("GITHUB_TOKEN")
 
-	// Each credential env var is checked individually so that a
-	// misconfigured CI secret surfaces as a clear test failure
-	// instead of silently skipping the related agent tests.
+	// All listed agent credentials must be set to run the suite.
 	for _, cfg := range agentConfigs {
-		if _, ok := os.LookupEnv(cfg.EnvVar); ok && *cfg.SecretValue == "" {
-			Fail(cfg.EnvVar + " is set but empty")
+		if *cfg.SecretValue == "" {
+			Fail(cfg.EnvVar + " must be set")
 		}
-	}
-	if oauthToken == "" && codexAuthJSON == "" {
-		Fail("No agent credentials set (CLAUDE_CODE_OAUTH_TOKEN, CODEX_AUTH_JSON)")
 	}
 })
--- a/test/e2e/task_test.go
+++ b/test/e2e/task_test.go
@@ -182,8 +182,13 @@ func describeAgentTests(cfg agentTestConfig) {
 			Expect(outputs).To(MatchRegexp(`input-tokens: \d+`))
 			Expect(outputs).To(MatchRegexp(`output-tokens: \d+`))
 
-			if cfg.AgentType == "claude-code" {
+			if cfg.SupportsCost {
 				Expect(outputs).To(MatchRegexp(`cost-usd: [\d.]+`))
+			}
+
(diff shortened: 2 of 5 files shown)


gjkim42 (commented):
can you add e2e tests? (we can add extra validation for the existing e2e test... maybe?)
and can we expand this to other agents as well? like codex, gemini...

knechtionscoding:
@gjkim42 working on an e2e test.

> and can we expand this to other agents as well? like codex, gemini...

Done

knechtionscoding:
@gjkim42 I think this is ready now. I added in capturing response from the other agents. It required a little more refactoring. Added E2E tests as well

test/e2e/suite_test.go
```diff
@@ -67,7 +96,16 @@ var _ = BeforeSuite(func() {
...
+	// At least one agent credential must be set to run the suite.
+	hasCredential := false
+	for _, cfg := range agentConfigs {
+		if *cfg.SecretValue != "" {
+			hasCredential = true
+			break
+		}
+	}
+	if !hasCredential {
+		Fail("No agent credentials set (CLAUDE_CODE_OAUTH_TOKEN, CODEX_AUTH_JSON, GEMINI_API_KEY, CURSOR_API_KEY)")
```
gjkim42: I'd like to support at least both claude code and codex. 
Unfortunately, I cannot cover others for now due to the budget.

remove gemini and cursor.
the tests should fail if any of the agents listed here doesn't exist

knechtionscoding: Done @gjkim42

test/e2e/task_test.go
```diff
@@ -195,9 +201,83 @@ func describeAgentTests(cfg agentTestConfig) {
...
+
+			if cfg.SupportsResponse {
+				By("verifying response is captured in Results as base64")
+				Expect(results).To(HaveKey("response"))
+				Expect(results["response"]).To(MatchRegexp(`^[A-Za-z0-9+/]+=*$`))
+			}
+		})
+	})
+
+	Describe(fmt.Sprintf("Slack response delivery [%s]", cfg.AgentType), func() {
```
gjkim42: can we remove this? I think this is a duplicate of the existing one.
(unless we can verify the slack actually gets a response...?)

knechtionscoding: Done. I had added it originally and then backtracked and updated the existing one and forgot to remove

PR #1318, opened 2026-06-08
feat: add WorkerPool CRD for persistent task execution

### LINKED ISSUE #911: API: Add persistent execution mode to TaskSpawner for warm agent sessions that eliminate per-task cold-start overhead

### LINKED ISSUE #1310: API: Add worker pool CRD for persistent Execution mode
In support of #911 we want to have a worker pool CRD that can be used to run persistent tasks from both a task spawner and individual tasks.

Should own the entire lifecycle of the statefulset and associated resources. Should be referencable as a workerPollRef on the task spec.

Should contain all fields of a statefulset configuration minus any reserved fields for specific container names

knechtionscoding:
#### What type of PR is this?

/kind feature

Adds a `WorkerPool` CRD that manages persistent worker pods for task execution, addressing the cold-start overhead problem identified in #911. Instead of creating a new Job per task (with git clone, dependency install, etc.), tasks can reference a `WorkerPool` and execute on pre-warmed pods backed by a StatefulSet with persistent storage.

**Architecture:**
- A `WorkerPool` owns a StatefulSet + headless Service. Each pod runs a `kelos-worker-runner` sidecar that polls for task assignments via pod annotations.
- The `WorkerPoolReconciler` handles both infrastructure (StatefulSet lifecycle, status) and task dispatch (assigning queued tasks to available pods, monitoring completion).
- Tasks opt in via `spec.workerPoolRef`. The existing Job-based path is unchanged for tasks without a pool reference.

**API changes:**
- New CRD: `WorkerPool` (spec: worker, replicas, volumeClaimTemplate using native `PersistentVolumeClaimSpec` for full PVC flexibility; status: phase, replicas, readyReplicas, conditions)
- `spec.worker` (WorkerSpec) promoted to canonical inline execution source for both pooled and non-pooled execution. Without `workerPoolRef`, creates a Job. With `workerPoolRef`, overrides pool defaults.
- Legacy flat fields (`type`, `credentials`, `model`, `effort`, `image`, `workspaceRef`, `agentConfigRefs`, `podOverrides`) deprecated with specific replacement references (e.g., "use spec.worker.type instead") but remain functional.
- `Credentials` changed from value to pointer (`*Credentials`) on TaskSpec/TaskTemplate so pooled tasks are admissible without providing credentials.
- `status.podName` unified for both Job and WorkerPool execution (no separate `workerPodName` field). Clients check one field regardless of execution backend.
- CEL validation enforces exactly one execution source: `worker`, `workerPoolRef`, or legacy `type+credentials` (mutually exclusive). Worker and legacy fields cannot be combined.
- Controller uses `resolveTask*()` helpers to read from `spec.worker` first, falling back to legacy fields for backward compatibility.

**Components added:**
- `api/v1alpha2/workerpool_types.go` — CRD types
- `internal/controller/workerpool_controller.go` — Reconciler (infrastructure + task dispatch)
- `cmd/kelos-worker-runner/` — Binary that runs in worker pods
- `internal/workerrunner/` — Worker runner core logic
- `internal/cli/get_workerpool.go` — `kelos get workerpools` command
- Helm CRD template, RBAC ClusterRole, Dockerfile
- Unit tests, integration test, example manifests
- `docs/reference.md` updated with WorkerSpec, WorkerPool, and deprecation annotations

Fixes #1310
Fixes #911

- **Backward compatibility**: All existing manifests continue to work unchanged. The deprecated top-level fields on TaskSpec/TaskTemplate are resolved via `resolveTask*()` helpers that every consumer (job builder, controller, CLI, metrics) uses.
- **`Credentials` field changed from value to pointer** on TaskSpec/TaskTemplate to make it optional when `workerPoolRef` is set. All accesses are nil-guarded.
- **Unified `status.podName`**: Both Job-based and WorkerPool-based tasks set the same `status.podName` field. No separate `workerPodName` — clients check one place.
- **`volumeClaimTemplate` uses native PVC spec**: Instead of a custom `WorkerPoolStorage` wrapper (size + storageClassName), we use `corev1.PersistentVolumeClaimSpec` directly so users get accessModes, volumeMode, dataSource, selector, etc. without future API changes.
- **Reconcile dispatch**: The `WorkerPoolReconciler` handles both WorkerPool objects and Tasks (with `workerPoolRef`). It disambiguates by trying WorkerPool first, falling through to Task only on NotFound.
- **Task assignment protocol**: Controller annotates pod with task name → runner detects annotation → runner executes agent → runner writes status annotation → controller reads status and updates Task phase → controller clears annotations.
- **No token refresh yet**: The worker runner receives `KELOS_TOKEN_SECRET` env var for future token refresh support, but the actual refresh logic is deferred.
- The `TaskReconciler` has an early return for tasks with `workerPoolRef` set, so it never creates Jobs for pool-based tasks.

```release-note
Add WorkerPool CRD for persistent task execution. WorkerPools manage pre-warmed worker pods via StatefulSet with native PersistentVolumeClaimSpec for storage. Tasks reference a pool via spec.workerPoolRef to execute on persistent infrastructure instead of creating one-shot Jobs. TaskSpec/TaskTemplate gain a spec.worker struct as the canonical inline execution source (type, credentials, image, model, workspaceRef, agentConfigRefs, podOverrides); legacy top-level fields are deprecated but remain functional. Task status.podName is unified for both Job and WorkerPool execution.
```


knechtionscoding:
@gjkim42 for your review. The vast majority of this diff is related to generating new CRDs. The rest is reworking #1228 into the new model.

I ran this through the general reviewer and API review locally twice and resolved open issues

knechtionscoding:
Due to the size, I am also happy to own running and posting of the reviewer if you would like and then just allow you to review the output to not burn the projects tokens

gjkim42 (commented):
I think we have to add e2e tests for the persistent tasks

knechtionscoding (commented):
## API Design Review

**Verdict**: APPROVE
**Scope**: New WorkerPool CRD + TaskSpec/TaskTemplate deprecation layer for execution-environment fields via worker struct

## Findings

### API Conventions
- **`workerpool_types.go:103` — Missing `ObservedGeneration` in `WorkerPoolStatus`**: Standard Kubernetes status convention includes `observedGeneration int64` so clients can determine if the controller has reconciled the latest spec. Many GitOps tools and dashboards check for this. Recommend adding `ObservedGeneration int64 \`json:"observedGeneration,omitempty"\`` — it's additive but best practice to include from day one.

### Compatibility
- **`TaskSpec.Credentials` value→pointer change is safe for v1alpha1**: The Go SDK change (value `Credentials` → `*Credentials`) breaks existing Go clients accessing the field without nil-checking, but this is acceptable given alpha versioning. The CRD schema change (required→optional) is a valid loosening — existing in-cluster objects decode without issue.
- **CEL union validation correctly permits all three forms**: The rule `has(self.workerPoolRef) || (worker w/ type+credentials) || (deprecated type+credentials)` covers existing manifests. Verified `examples/01-simple-task/task.yaml` and `self-development/` manifests pass the third disjunct.
- **No tightening of previously-accepted values**: Existing manifests in `self-development/` and `examples/` continue to validate unchanged. Good.

### Naming & Consistency
- **`WorkerPoolReference.Name` has `+kubebuilder:validation:MinLength=1` but `WorkspaceReference.Name` does not** (`workspace_types.go:106`): Minor inconsistency. The MinLength=1 is good practice (prevents empty strings), but the discrepancy could confuse users. Non-blocking — consider adding MinLength=1 to WorkspaceReference in a follow-up.

### Validation
- **No mutual exclusion between `worker.*` and deprecated top-level fields**: A user can set both `spec.type: claude-code` and `spec.worker.type: codex`. The `Effective*()` accessors resolve correctly (worker wins), but the dual-specification may confuse users. Acceptable for a deprecation transition period — consider adding a CEL warning or webhook admission warning in a follow-up.
- **WorkerConfig credential XValidation rule is sound**: `!has(self.credentials) || self.credentials.type == 'none' || has(self.credentials.secretRef)` correctly handles the optional-when-pool case.

### Extensibility
- **`WorkerPoolStorage` is appropriately minimal** (just `size` + `storageClassName`): Good API minimality. PVC access modes and volume mode can be added later without breaking changes.
- **`WorkerPoolSpec` reuses existing Kelos types** (Credentials, WorkspaceReference, AgentConfigReference, PodOverrides): Consistent with the existing API surface, no synonym introduction.

## Suggestions (optional)
- Add `ObservedGeneration` to WorkerPoolStatus before GA — it's costless now and painful to add to existing controllers later when clients already depend on its absence.
- Consider a CEL validation that prevents setting `worker` fields AND deprecated top-level fields simultaneously (or at least documents precedence in the CRD description). Currently documented only in godoc.
- The `TaskWorker = WorkerConfig` / `TaskTemplateWorker = WorkerConfig` type aliases work correctly for code generation but their godoc ("is an alias for WorkerConfig used on TaskSpec") may not surface in API reference docs. Consider verifying how these render in the generated CRD reference.

/kelos needs-input

internal/controller/workerpool_controller.go
```diff
@@ -0,0 +1,1202 @@
...
+		task.Status.WorkerPodName = availablePod.Name
+		now := metav1.Now()
+		task.Status.StartTime = &now
+		return r.Status().Update(ctx, task)
+	}); err != nil {
+		return ctrl.Result{}, fmt.Errorf("workerpool %s: updating task %s status for assignment: %w", poolName, task.Name, err)
+	}
+
+	// Annotate the pod with the assigned task
+	podPatch := client.MergeFrom(availablePod.DeepCopy())
```
knechtionscoding: [P1] `client.MergeFrom` does not include a resourceVersion check, so two concurrent reconciles that both select the same available pod will both succeed their patch — the second silently overwrites the first's `AnnotationWorkerAssignedTask`. The runner picks up only the last writer's task while the first task's status permanently claims this pod.

Fix: use `client.MergeFromWithOptions(availablePod.DeepCopy(), client.MergeFromWithOptimisticLock{})` so the second patch returns a conflict error and triggers the existing rollback + requeue path.

internal/taskbuilder/builder.go
```diff
@@ -109,6 +109,9 @@ func (tb *TaskBuilder) BuildTask(
 	if taskTemplate.UpstreamRepo != "" {
 		task.Spec.UpstreamRepo = taskTemplate.UpstreamRepo
 	}
+	if taskTemplate.WorkerPoolRef != nil {
```
knechtionscoding, line 112: [P1] `WorkerPoolRef` is propagated here, but the new `taskTemplate.Worker` field is not. If a TaskSpawner uses `taskTemplate.worker` (the non-deprecated path), spawned tasks will have no type/credentials and will either fail CEL validation or produce empty-config Jobs.

```suggestion
	if taskTemplate.WorkerPoolRef != nil {
		task.Spec.WorkerPoolRef = taskTemplate.WorkerPoolRef
	}
	if taskTemplate.Worker != nil {
		task.Spec.Worker = taskTemplate.Worker
	}
```

knechtionscoding (commented):
## Review Summary

**Verdict**: REQUEST CHANGES
**Overall correctness**: patch is incorrect
**Scope**: Adds WorkerPool CRD, controller, worker runner binary, CLI support, and TaskSpec deprecation layer for persistent task execution

## Findings Overview

| P0       | 0     | —         | none |
| P1       | 2     | internal/controller/workerpool_controller.go:855, internal/taskbuilder/builder.go:66-114 | Race in task assignment + Worker field not propagated |
| P2       | 2     | docs/reference.md, internal/workerrunner/runner.go:55 | Missing docs + silent config parse failure |
| P3       | 0     | —         | none |

## Findings

### Correctness
- [P1] internal/controller/workerpool_controller.go:855 — The pod annotation patch in `assignTask` uses `client.MergeFrom` which does not include a resourceVersion check. If two tasks reconcile simultaneously and both select the same available pod (stale List), both merge patches will succeed — the second overwrites the first annotation value. Task A's status claims the pod but the runner only sees Task B. Fix: use `client.MergeFromWithOptions(availablePod.DeepCopy(), client.MergeFromWithOptimisticLock{})` so that the second patch fails with a conflict and triggers the existing rollback path.
- [P1] internal/taskbuilder/builder.go:66-114 — `TaskBuilder.BuildTask` copies `taskTemplate.Type`, `taskTemplate.Credentials`, and other deprecated top-level fields to the spawned Task, but does NOT propagate `taskTemplate.Worker`. If a TaskSpawner uses the new `worker` struct (as the deprecation docs encourage), spawned tasks will have empty type/credentials and fail CEL validation or produce incorrect Jobs. Add `if taskTemplate.Worker != nil { task.Spec.Worker = taskTemplate.Worker }` alongside the existing WorkerPoolRef propagation.

### Documentation
- [P2] docs/reference.md — This PR introduces a new CRD (WorkerPool), a new CLI command (`kelos get workerpools`), new flags (`--worker-runner-image`, `--worker-runner-image-pull-policy`), and new API fields (`TaskSpec.worker`, `TaskSpec.workerPoolRef`, `TaskStatus.workerPodName`, `TaskTemplate.worker`, `TaskTemplate.workerPoolRef`). The canonical reference docs were not updated. Per project conventions, user-facing changes require matching docs in the same PR.

### Conventions
- [P2] internal/workerrunner/runner.go:55 — `ConfigFromEnv` silently ignores parse errors for `KELOS_IDLE_TIMEOUT` and `KELOS_MAX_TASKS_PER_WORKER`, falling back to defaults. Per project conventions (fail-fast on invalid configuration), invalid duration or integer values should log a warning or return an error rather than silently degrading to the default. An operator debugging why their 5s idle timeout is not taking effect will not get any signal.

## Suggestions
- [P3] Consider adding `ObservedGeneration` to `WorkerPoolStatus` before GA — standard Kubernetes status convention; additive now, harder to add later.
- [P3] The `assignTask` function selects the first available pod from the list. Consider adding a scheduling heuristic (e.g., fewest tasks completed) to distribute work more evenly across workers.

## Key takeaways
- The optimistic-lock-less MergeFrom patch on pod assignment is a real correctness issue under concurrent load — one-line fix.
- The `Worker` field not propagating through the task builder breaks the new API path for TaskSpawners — also a straightforward fix.
- Recommend running `/kelos api-review` for deeper API design analysis of the new CRD fields.

internal/workerrunner/runner.go
```diff
@@ -0,0 +1,285 @@
...
+	IdleTimeout       time.Duration
+	MaxTasksPerWorker int32
+}
+
+// ConfigFromEnv reads the worker runner configuration from environment variables.
+func ConfigFromEnv() Config {
+	cfg := Config{
+		PodName:      os.Getenv("KELOS_POD_NAME"),
+		PodNamespace: os.Getenv("KELOS_POD_NAMESPACE"),
+		AgentType:    os.Getenv("KELOS_AGENT_TYPE"),
```
knechtionscoding, line 55: [P2] Parse errors for `KELOS_IDLE_TIMEOUT` and `KELOS_MAX_TASKS_PER_WORKER` are silently swallowed, falling back to defaults. Per project conventions (fail-fast on invalid configuration), log a warning or fail startup so operators get a signal when their config is malformed.

internal/controller/workerpool_controller.go
knechtionscoding: fixed in: e9ce8348c45c354a206746bb2a680c99fb65a006

internal/taskbuilder/builder.go
knechtionscoding: fixed in: fixed in: e9ce8348c45c354a206746bb2a680c99fb65a006

internal/workerrunner/runner.go
knechtionscoding: fixed in: e9ce8348c45c354a206746bb2a680c99fb65a006

knechtionscoding (commented):
## API Design Review

**Verdict**: APPROVE
**Scope**: New WorkerPool CRD + WorkerConfig shared type + TaskSpec/TaskTemplate deprecation layer with worker struct and workerPoolRef

## Findings

### API Conventions ✓
- All field names correctly use camelCase in Go structs and JSON tags
- Optional fields properly marked with `+optional`, pointer types, and `omitempty`
- Required fields validated with `+kubebuilder:validation:Required`
- No boolean `isX`/`enableX` anti-patterns
- `WorkerPoolPhase` correctly uses named string type with const block
- Status fields (`Replicas`, `ReadyReplicas`, `Phase`) appropriately mirror standard Kubernetes status patterns (matching Deployment/StatefulSet conventions)
- `ObservedGeneration` included in `WorkerPoolStatus` — good practice for day-one

### Compatibility ✓
- **No breaking changes**: deprecated top-level fields remain functional with `+deprecated` markers, `+optional`, pointer types, and `omitempty`
- **CEL union validation** correctly accepts all three forms: `workerPoolRef`, `worker.type+credentials`, or deprecated `type+credentials`
- `Credentials` value→pointer change is a valid loosening for v1alpha1 (required→optional in CRD schema)
- Existing manifests in `examples/01-simple-task/`, `self-development/` continue to validate under the updated schema
- `TaskSpawnerSpec` workspace validation rule correctly extended to cover `taskTemplate.worker.workspaceRef` and `taskTemplate.workerPoolRef`
- `TaskBuilder` correctly propagates both `Worker` and `WorkerPoolRef` from templates (lines 112-117)

### Naming & Consistency ✓
- `WorkerPoolReference` follows existing `XxxReference` pattern (`WorkspaceReference`, `AgentConfigReference`, `SecretReference`)
- Field naming (`workerPoolRef`, `workspaceRef`, `agentConfigRef`) follows established `xxxRef` convention
- `replicas`, `readyReplicas`, `phase`, `conditions`, `observedGeneration` all match upstream Kubernetes semantics
- `WorkerConfig` correctly reused via type aliases for `TaskWorker` and `TaskTemplateWorker`

### Validation ✓
- `WorkerPoolSpec` credential XValidation rule matches existing pattern from previous `TaskSpec`
- `agentConfigRef`/`agentConfigRefs` mutual exclusion enforced at all three levels (WorkerPoolSpec, WorkerConfig, TaskSpec)
- `WorkerPoolReference.Name` has `+kubebuilder:validation:MinLength=1` — good empty-string prevention (better than existing `WorkspaceReference` which lacks it)
- `WorkerPoolStorage.Size` correctly uses `resource.Quantity` type

### Extensibility ✓
- `WorkerPoolStorage` struct (rather than bare `size` scalar on spec) allows future fields (access modes, volume mode) without breaking changes
- `WorkerPoolSpec` uses `PodOverrides` pointer — expandable without schema change
- `WorkerPoolReference` is a struct (not bare string) — can add `namespace` later for cross-namespace refs

### Documentation ✓
- `docs/reference.md` updated with WorkerPool section, deprecated field markers, new CLI commands
- Example manifests under `examples/15-workerpool/` cover all three usage patterns (pool, task-with-pool, spawner-with-pool)
- All exported types and fields have godoc comments

## Suggestions (non-blocking)

- **WorkerPoolSpec field duplication vs WorkerConfig embedding**: `WorkerPoolSpec` duplicates fields (type, credentials, model, etc.) that `WorkerConfig` already groups. Consider embedding `WorkerConfig` in `WorkerPoolSpec` with different validation markers in a follow-up to reduce maintenance surface — though the required-vs-optional distinction makes direct embedding non-trivial, so the current approach is reasonable.
- **`WorkspaceReference` and `AgentConfigReference` lack `MinLength=1`**: The new `WorkerPoolReference` correctly adds it. Consider adding `MinLength=1` to the older reference types in a follow-up for consistency.
- **No CEL warning when both `worker.*` and deprecated top-level fields are set simultaneously**: The `Effective*()` accessors resolve correctly (worker wins), but dual-specification may confuse users. A webhook admission warning could surface this during the deprecation transition.
- **`taskspawner-with-pool.yaml` includes `workspaceRef` alongside `workerPoolRef`**: The comment explains why (spawner discovery loop). Consider noting in the reference docs that spawners with pool-based templates still need `workspaceRef` for GitHub/Linear source discovery, since the CEL rule on `TaskSpawnerSpec` accepts `workerPoolRef` alone but the spawner's runtime behavior may differ.

/kelos needs-input

knechtionscoding (commented):
## Review Summary

**Verdict**: COMMENT
**Scope**: Adds WorkerPool CRD, controller, worker-runner binary, CLI support, Helm RBAC, TaskSpec deprecation layer via worker struct, and comprehensive tests

## Findings Overview

| P0       | 0     | — | none |
| P1       | 0     | — | none |
| P2       | 2     | See below | DependsOn/branch-lock not enforced for pool tasks; kelos logs does not resolve WorkerPodName |
| P3       | 1     | See below | clearPodAssignment uses MergeFrom without optimistic lock |

## Findings

### Correctness

- [P2] `internal/controller/workerpool_controller.go:807` — The `TaskReconciler` early-returns for tasks with `workerPoolRef` (task_controller.go:80), skipping `checkDependencies` and `BranchLocker.TryAcquire`. The `WorkerPoolReconciler.reconcileTask` has no equivalent logic — it immediately calls `assignTask` for any non-terminal task via the `default` case. A Task with both `workerPoolRef` and `dependsOn` (valid per the schema) will execute immediately, ignoring its dependencies. Similarly, branch locking is bypassed. If `dependsOn` is intentionally unsupported for pool-based tasks, consider adding CEL validation to reject it or documenting the limitation. Otherwise, port the dependency-check logic (or at minimum a `TaskPhaseWaiting` guard that skips assignment when the task is in Waiting phase).

- [P3] `internal/controller/workerpool_controller.go:954` — `clearPodAssignment` uses `client.MergeFrom` without `MergeFromWithOptimisticLock`. Lower risk than `assignTask` (since only the controller clears annotations after completion), but a concurrent runner `setTaskStatus` call could theoretically lead to a lost update on the `tasks-completed` counter. Acceptable for now given annotation semantics.

### CLI

- [P2] `internal/cli/logs.go:131-144` — `resolveTaskPodName` searches for pods labeled `kelos.dev/task` and falls back to `task.Status.PodName`. Worker pool pods are not labeled with the task name (assignment is via annotation), so `kelos logs <task>` will fail to find the pod for pool-based tasks. The function should also check `task.Status.WorkerPodName` as a fallback:
```go
if task.Status.WorkerPodName != "" {
    return task.Status.WorkerPodName, nil
}
```

## Suggestions

- [P3] Consider adding an explicit `case kelosv1alpha1.TaskPhaseWaiting` arm in `reconcileTask` rather than falling through to the `default` case — this documents intent and prevents future regressions if dependency logic is added later.
- [P3] The previous review's P1 findings (optimistic lock on `assignTask`, Worker field propagation in task builder) have both been addressed — confirmed at lines 856 and 112 respectively.

## Key takeaways

- The core architecture (StatefulSet lifecycle, annotation-based task protocol, optimistic-lock assignment, rollback on patch failure) is solid and well-tested with unit, integration, and e2e coverage.
- Main gap is dependency/branch-lock handling for pool-based tasks — straightforward to either explicitly forbid via CEL or implement in a follow-up.
- `kelos logs` needs a one-line fix to support pool-based tasks.

knechtionscoding:
@gjkim42 ready for review.

knechtionscoding:
@gjkim42 bump :)

gjkim42:
Will take a look this weekend.

api/v1alpha1/task_types.go
```diff
@@ -195,23 +195,25 @@ type PodOverrides struct {
 	ExtraInitContainers []corev1.Container `json:"extraInitContainers,omitempty"`
 }
 
-// TaskSpec defines the desired state of Task.
+// WorkerConfig groups execution environment fields shared by Task and
```
gjkim42: Blocking API shape: I think this should be a public `WorkerSpec`, not `WorkerConfig` / `TaskWorker`. The reusable API concept is worker execution spec, and `Task.spec.worker`, `TaskTemplate.worker`, and `WorkerPool.spec.worker` can all use that same shape: type, credentials, workspaceRef, model, effort, image, agentConfig*, and podOverrides. Since requiredness differs by parent, keep the `WorkerSpec` fields optional and enforce parent-specific requiredness with CEL first.

api/v1alpha1/workerpool_types.go
```diff
@@ -0,0 +1,181 @@
...
+	// +kubebuilder:validation:Required
+	Size resource.Quantity `json:"size"`
+
+	// StorageClassName is the name of the StorageClass to use.
+	// When empty, the cluster default StorageClass is used.
+	// +optional
+	StorageClassName *string `json:"storageClassName,omitempty"`
+}
+
+// WorkerPoolSpec defines the desired state of WorkerPool.
```
gjkim42, line 52: Blocking API shape: `WorkerPoolSpec` should probably contain `WorkerSpec` instead of duplicating worker fields at the top level. Suggested shape: `WorkerPoolSpec` has `Worker WorkerSpec`, `Replicas int32`, and `Storage WorkerPoolStorage`. Then add parent-level CEL on `WorkerPoolSpec` requiring `worker.type`, `worker.credentials`, and `worker.workspaceRef`. A webhook can later handle cross-resource compatibility, but this local requiredness should be enforced by CEL now.

api/v1alpha1/task_types.go
```diff
@@ -244,14 +248,95 @@ type TaskSpec struct {
...
+type TaskWorker = WorkerConfig
+
+// TaskSpec defines the desired state of Task.
+//
+// Exactly one of workerPoolRef or worker.type+worker.credentials must be
+// provided. When workerPoolRef is set, the Task executes on a pre-warmed
+// worker pod; the pool supplies type, credentials, and workspace. When
+// worker is set with type and credentials, a one-shot Job is created.
+//
+// +kubebuilder:validation:XValidation:rule="has(self.workerPoolRef) || (has(self.worker) && has(self.worker.type) && size(self.worker.type) > 0 && has(self.worker.credentials)) || (has(self.type) && size(self.type) > 0 && has(self.credentials))",message="either workerPoolRef, worker with type and credentials, or type and credentials (deprecated) is required"
```
gjkim42: Blocking validation contract: this should enforce the execution source choice with CEL now. For `Task`, I think the rule should require exactly one of inline `worker`, `workerPoolRef`, or deprecated top-level `type` + `credentials`. While per-task pool overrides are not implemented, please reject `workerPoolRef` together with inline `worker` or deprecated top-level execution fields. When inline `worker` is used, require non-empty `worker.type` and `worker.credentials`.

api/v1alpha1/taskspawner_types.go
```diff
@@ -828,52 +828,70 @@ type TaskTemplateMetadata struct {
...
+// TaskTemplateWorker is an alias for WorkerConfig used on TaskTemplate.
+type TaskTemplateWorker = WorkerConfig
+
 // TaskTemplate defines the template for spawned Tasks.
 //
+// Exactly one of workerPoolRef or worker.type+worker.credentials must be
+// provided. When workerPoolRef is set, spawned Tasks execute on pre-warmed
+// workers; the pool supplies type, credentials, and workspace.
+//
+// +kubebuilder:validation:XValidation:rule="has(self.workerPoolRef) || (has(self.worker) && has(self.worker.type) && size(self.worker.type) > 0 && has(self.worker.credentials)) || (has(self.type) && size(self.type) > 0 && has(self.credentials))",message="either workerPoolRef, worker with type and credentials, or type and credentials (deprecated) is required"
```
gjkim42: Same API contract issue as `Task`: `TaskTemplate` should enforce exactly one execution source with CEL. Please require one of inline `worker`, `workerPoolRef`, or deprecated top-level `type` + `credentials`; reject mixed `workerPoolRef` + inline/deprecated worker config until pool overrides are implemented; and require `worker.type` + `worker.credentials` when inline `worker` is used.

api/v1alpha1/taskspawner_types.go
```diff
@@ -943,7 +962,7 @@ type TaskTemplate struct {
 }
 
 // TaskSpawnerSpec defines the desired state of TaskSpawner.
-// +kubebuilder:validation:XValidation:rule="!(has(self.when.githubIssues) || has(self.when.githubPullRequests) || has(self.when.githubWebhook) || has(self.when.linearWebhook)) || has(self.taskTemplate.workspaceRef)",message="taskTemplate.workspaceRef is required when using githubIssues, githubPullRequests, githubWebhook, or linearWebhook source"
+// +kubebuilder:validation:XValidation:rule="!(has(self.when.githubIssues) || has(self.when.githubPullRequests) || has(self.when.githubWebhook) || has(self.when.linearWebhook)) || has(self.taskTemplate.workspaceRef) || (has(self.taskTemplate.worker) && has(self.taskTemplate.worker.workspaceRef)) || has(self.taskTemplate.workerPoolRef)",message="taskTemplate.workspaceRef, taskTemplate.worker.workspaceRef, or taskTemplate.workerPoolRef is required when using githubIssues, githubPullRequests, githubWebhook, or linearWebhook source"
```
gjkim42: Blocking contract issue: `taskTemplate.workerPoolRef` should not satisfy the GitHub/Linear source workspace requirement unless the controller actually resolves the referenced WorkerPool first. CEL cannot dereference `workerPoolRef`, and the current controller path only resolves top-level `taskTemplate.workspaceRef` when building the spawner pod. For now, either keep requiring top-level `taskTemplate.workspaceRef` for these sources, or implement effective workspace resolution from `taskTemplate.worker.workspaceRef` / `WorkerPool.spec.worker.workspaceRef` before relaxing this validation.

gjkim42:
a Task should use either WorkerSpec or WorkerPoolRef.

knechtionscoding:
@gjkim42 Updated based on your review comments and aligned these into the v1alpha2 versions

gjkim42 (commented):
Let's not update v1alpha1 and mark the old fields as deprecated.

gjkim42 (commented):
s/Copyright 2025/Copyright 2026

knechtionscoding:
@gjkim42 Updated based on your review comments and aligned these into the v1alpha2 versions

api/v1alpha2/task_types.go
```diff
@@ -204,20 +204,83 @@ type PodOverrides struct {
 	ExtraInitContainers []corev1.Container `json:"extraInitContainers,omitempty"`
 }
 
+// WorkerSpec groups execution environment fields shared by Task,
+// TaskTemplate, and WorkerPool. When workerPoolRef is set on a Task or
+// TaskTemplate, these fields are optional and default to the pool's
+// configuration.
+//
+// +kubebuilder:validation:XValidation:rule="!has(self.credentials) || self.credentials.type == 'none' || has(self.credentials.secretRef)",message="credentials.secretRef is required for api-key and oauth credential types"
+type WorkerSpec struct {
```
gjkim42, line 213: can we mark the same fields from task.spec as deprecated?

gjkim42 (commented):
/kelos api-review
xref: https://github.com/kelos-dev/kelos/issues/911#issuecomment-4638287504

api/v1alpha2/task_types.go
knechtionscoding: fixed

knechtionscoding:
@gjkim42 Bump for feedback.

api/v1alpha2/task_types.go
```diff
@@ -322,6 +401,11 @@ type TaskStatus struct {
 	// Results contains structured key-value outputs produced by the agent.
 	// +optional
 	Results map[string]string `json:"results,omitempty"`
+
+	// WorkerPodName is the name of the worker pod executing this Task.
+	// Set only when the Task references a WorkerPool.
+	// +optional
+	WorkerPodName string `json:"workerPodName,omitempty"`
```
gjkim42: why don't we just use the existing podName?

api/v1alpha2/workerpool_types.go
```diff
@@ -0,0 +1,166 @@
...
+	Worker WorkerSpec `json:"worker"`
+
+	// Replicas is the desired number of persistent worker pods.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=0
+	Replicas int32 `json:"replicas"`
+
+	// Storage configures persistent storage for worker pods.
+	// +kubebuilder:validation:Required
+	Storage WorkerPoolStorage `json:"storage"`
```
gjkim42: Why don't we use VolumeClaimTemplate of kubernetes instead of making our own?

gjkim42 (commented):
/kelos api-review

knechtionscoding:
@gjkim resolved review comments

test/e2e/workerpool_test.go
```diff
@@ -0,0 +1,336 @@
...
+	kelos "github.com/kelos-dev/kelos/api/v1alpha2"
+	"github.com/kelos-dev/kelos/test/e2e/framework"
+)
+
+var _ = Describe("WorkerPool", func() {
+	f := framework.NewFramework("workerpool")
+
+	BeforeEach(func() {
+		if oauthToken == "" {
+			Skip("CLAUDE_CODE_OAUTH_TOKEN not set, skipping WorkerPool e2e tests")
```
gjkim42: Some e2e test files have this convention but I think we have to remove all...

WorkerPool is a core concept, so we should not skip e2e tests.

test/e2e/workerpool_test.go
```diff
@@ -0,0 +1,336 @@
...
+		f.WaitForTaskPhase("pool-task-2", "Succeeded")
+	})
+})
+
+var _ = Describe("WorkerPool with TaskSpawner", func() {
+	f := framework.NewFramework("wp-spawner")
+
+	BeforeEach(func() {
+		if githubToken == "" || oauthToken == "" {
+			Skip("GITHUB_TOKEN or CLAUDE_CODE_OAUTH_TOKEN not set, skipping WorkerPool+TaskSpawner e2e tests")
```
gjkim42: I think we need to remove. this as well

test/e2e/workerpool_test.go
```diff
@@ -0,0 +1,336 @@
+package e2e
```
gjkim42, line 1: 👍

gjkim42 (commented):
Thanks for the core contribution!
API design and e2e tests look good.

/kelos review

knechtionscoding:
@gjkim42 Resolved comments/review and ran it through the reviewer a few more times, both api and regular reviewer, locally.

knechtionscoding:
@gjkim42 fixed and re-ran reviewer. Came back clean for me.

gjkim42:
@knechtionscoding 
I guess the new e2e tests failed

knechtionscoding:
> I guess the new e2e tests failed

@gjkim42 I believe this is because of infra in the E2E cluster. the pod is stuck in pending (likely a PVC)? Or rather it can't be scheduled. Unsure if we can get logs from somewhere to show that?

knechtionscoding:
Closing in favor of #1415

PR #1328, opened 2026-06-10
Fix/plugin configmap delivery

scottt732:
#### What type of PR is this?

/kind bug

The `plugin-setup` init container inlines every plugin skill's and agent's
content base64-encoded into a single `sh -c <script>` command string. Linux
caps a single `execve` argument at 128KiB (`MAX_ARG_STRLEN`), so any
AgentConfig whose plugin content exceeds that fails at container start with:

    exec /bin/sh: argument list too long

(exit 255, before the agent ever runs — observed in production with a
reviewer-attachment-heavy AgentConfig).

This PR delivers plugin content via a per-task ConfigMap instead:

- `buildPluginConfigMapData` produces a `<taskName>-plugins` ConfigMap with
  one flat key per content blob (`p<i>-s<j>` / `p<i>-a<j>`) plus `KeyToPath`
  items mapping each key to the same layout the old script produced
  (`<plugin>/skills/<skill>/SKILL.md`, `<plugin>/agents/<agent>.md`).
  Existing `sanitizeComponentName` validation is retained.
- The Job mounts the ConfigMap read-only at `/kelos-plugin-src` in
  `plugin-setup`, whose script is now constant-size: it copies the staged
  tree into the writable plugin emptyDir.
- The Task reconciler creates/updates the ConfigMap the same way it manages
  the per-task github-token Secret: controller ownerReference for GC,
  idempotent conflict-retried update on reconcile, and a
  `PluginConfigMapFailed` warning event on failure.
- Deterministic build failures (invalid component names, payload over the
  ~900KiB budget) transition the Task to `Failed` with a status message,
  mirroring the `JobBuildFailed` path; API-level errors remain transient
  requeues.

N/A

- The copy script uses `cp -RL` and excludes `..*` entries because the
  kubelet atomic-writer materializes ConfigMap items as symlinks through
  hidden `..data`/`..<timestamp>` directories.
- `kelos-plugin-src` is added to `reservedVolumeNames` so podOverrides can't
  collide with the staging volume.
- RBAC: `configmaps get;list;watch;create;update` added to the
  kubebuilder markers; the chart role was regenerated via `make update`.
- Repro for the bug: any AgentConfig whose combined plugin skill/agent
  content exceeds ~96KiB raw (128KiB after base64 expansion) — `plugin-setup`
  exits 255 before the agent container starts.
- Tests: `TestBuildJob_AgentConfigPlugins` rewritten for the new shape; new
  `TestBuildPluginConfigMap`, `TestBuildPluginConfigMap_InvalidNames`,
  `TestBuildPluginConfigMap_TooLarge`,
  `TestEnsurePluginConfigMap_CreateAndUpdate`. `go build ./...`,
  `go vet ./...`, and full `make test` pass.

```release-note
Plugin content for Task pods is now delivered via a per-task ConfigMap (`<taskName>-plugins`) mounted into the plugin-setup init container, instead of being inlined base64-encoded into the init container command. This fixes `exec /bin/sh: argument list too long` failures for AgentConfigs with large plugin content. Plugin payloads over ~900KiB are rejected with a clear error and the Task transitions to Failed. The controller RBAC now requires get/list/watch/create/update on ConfigMaps - re-apply the chart/manifests when upgrading.
```

### DIFF
--- a/internal/controller/task_controller.go
+++ b/internal/controller/task_controller.go
@@ -83,7 +83,7 @@ type TaskReconciler struct {
 // +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=pods/log,verbs=get
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update
-// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 
 // Reconcile handles Task reconciliation.
@@ -351,6 +351,36 @@ func (r *TaskReconciler) createJob(ctx context.Context, task *kelos.Task) (ctrl.
 		return ctrl.Result{}, err
 	}
 
+	// Plugin content is delivered via a per-task ConfigMap that the Job's
+	// plugin-setup init container mounts, so it must exist before the Job.
+	if agentConfig != nil && len(agentConfig.Plugins) > 0 {
+		configMap, err := buildPluginConfigMap(task, agentConfig.Plugins)
+		if err != nil {
+			// Deterministic (invalid component names, payload over budget) —
+			// retrying can't succeed, so fail the Task like JobBuildFailed.
+			logger.Error(err, "unable to build plugin ConfigMap")
+			r.recordEvent(task, corev1.EventTypeWarning, "PluginConfigMapFailed", "Failed to build plugin ConfigMap: %v", err)
+			updateErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				if getErr := r.Get(ctx, client.ObjectKeyFromObject(task), task); getErr != nil {
+					return getErr
+				}
+				task.Status.Phase = kelos.TaskPhaseFailed
+				task.Status.Message = fmt.Sprintf("Failed to build plugin ConfigMap: %v", err)
+				return r.Status().Update(ctx, task)
+			})
+			if updateErr != nil {
+				logger.Error(updateErr, "Unable to update Task status")
+			}
+			return ctrl.Result{}, err
+		}
+		if err := r.ensurePluginConfigMap(ctx, task, configMap); err != nil {
+			// API-level failures are transient — requeue without failing the Task.
+			logger.Error(err, "unable to ensure plugin ConfigMap")
+			r.recordEvent(task, corev1.EventTypeWarning, "PluginConfigMapFailed", "Failed to ensure plugin ConfigMap: %v", err)
+			return ctrl.Result{}, err
+		}
+	}
+
 	// Set owner reference
 	if err := controllerutil.SetControllerReference(task, job, r.Scheme); err != nil {
 		logger.Error(err, "unable to set owner reference")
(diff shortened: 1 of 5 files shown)


internal/controller/task_controller.go
```diff
@@ -306,6 +307,16 @@ func (r *TaskReconciler) createJob(ctx context.Context, task *kelosv1alpha1.Task
 		return ctrl.Result{}, err
 	}
 
+	// Plugin content is delivered via a per-task ConfigMap that the Job's
+	// plugin-setup init container mounts, so it must exist before the Job.
+	if agentConfig != nil && len(agentConfig.Plugins) > 0 {
```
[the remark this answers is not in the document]
scottt732: Addressed in f37e34e: buildPluginConfigMap now runs in createJob, and its deterministic failures (invalid component names, payload over the ~900KiB budget) transition the Task to Failed with a status message — mirroring the JobBuildFailed path. ensurePluginConfigMap takes the prebuilt ConfigMap and only performs the API operations, whose errors remain transient requeues.

internal/controller/task_controller.go
```diff
@@ -427,6 +458,48 @@ func (r *TaskReconciler) resolveGitHubAppToken(ctx context.Context, task *kelosv
...
+		// Re-assert the controller reference too: an existing ConfigMap with
+		// a stripped ownerRef would otherwise outlive its Task. Concurrent
+		// reconciles can race here, so retry on conflict like the other
+		// update paths in this controller.
+		updateErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			existing := &corev1.ConfigMap{}
+			if err := r.Get(ctx, client.ObjectKey{Name: configMap.Name, Namespace: configMap.Namespace}, existing); err != nil {
+				return err
+			}
+			existing.Data = configMap.Data
```
gjkim42, line 604: [P0] This `AlreadyExists` path can adopt and overwrite an unrelated ConfigMap named `<task>-plugins`: it replaces all `Data`, then sets the Task as controller owner, so a user-created ConfigMap with the generated name can be modified and later garbage-collected with the Task. Please only update an existing ConfigMap after verifying it is already controlled by this Task, for example with `metav1.IsControlledBy(existing, task)`. If it is unowned or controlled by something else, return a name-collision error instead. A test with a pre-existing unowned ConfigMap would lock this down.

gjkim42 (changes requested):
API design review: one blocking finding. I omitted the broader `kelos-*` volume-name reservation topic because it is now tracked separately in #1343.

gjkim42:
Thanks for finding a critical bug.
Let me know if you fix the resource adoption.

@scottt732

internal/controller/task_controller.go
scottt732: Fixed in 4480032: the AlreadyExists update path now checks metav1.IsControlledBy(existing, task) and returns a name-collision error instead of adopting/overwriting a ConfigMap it does not control. Added TestEnsurePluginConfigMap_NotAdoptedWhenUnowned, which pre-creates an unowned <task>-plugins ConfigMap and asserts its data is preserved, no plugin content is written, and no owner reference is added.

scottt732:
@gjkim42 should be fixed now.

internal/controller/task_controller.go
```diff
@@ -60,7 +60,7 @@ type TaskReconciler struct {
 // +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=pods/log,verbs=get
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update
-// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;delete
```
gjkim42: we don't need delete here I guess.

gjkim42 (approved):
looks good. can you fix one minor issue?

then please squash commits and update the release note as well.

gjkim42:
@scottt732 can you fix those nits? I'll merge it after that

gjkim42:
Blockers
- squash commits
- Remove unnecessary delete permission

PR #1329, opened 2026-06-10
feat: add Task-native budget enforcement via TaskBudget and TaskRecord CRDs

### LINKED ISSUE #977: API: Add budgetPolicy to TaskSpawnerSpec for cost-aware governance with per-period and lifetime spending limits

## Summary

Add a `budgetPolicy` field to `TaskSpawnerSpec` that enforces cost ceilings on AI agent spending. Currently, Kelos captures per-task cost data (`cost-usd`, `input-tokens`, `output-tokens`) and emits Prometheus counters, but this data never flows back into the control plane to enforce limits. The existing `maxConcurrency` and `maxTotalTasks` govern task *count*, but task cost varies widely ($0.10 for a simple lint fix vs $15+ for a complex multi-file refactor), making count-based limits insufficient for budget management. This proposal adds cost-aware governance alongside the existing count-based fields, letting operators set per-period (daily/weekly/monthly) and lifetime cost ceilings per TaskSpawner, with automatic suspension when limits are reached.

### 1. Cost data is captured but never used for governance

The controller already captures per-task cost at `internal/controller/task_controller.go:888-910`:

```go
func RecordCostTokenMetrics(task *kelosv1alpha1.Task, results map[string]string) {
    // ...
    if costStr, ok := results["cost-usd"]; ok {
        if cost, err := strconv.ParseFloat(costStr, 64); err == nil && cost > 0 {
            taskCostUSD.WithLabelValues(labels...).Add(cost)
        }
    }
}
```

This feeds Prometheus counters (`kelos_task_cost_usd_total`, `kelos_task_input_tokens_total`, `kelos_task_output_tokens_total` — defined in `internal/controller/metrics.go:47-72`), but nothing reads these values back to make admission decisions. Cost observability exists; cost *control* does not.

### 2. Count-based limits are a poor proxy for cost

The existing governance fields on `TaskSpawnerSpec`:

```go
MaxConcurrency *int32 `json:"maxConcurrency,omitempty"`  // concurrent running tasks
MaxTotalTasks  *int32 `json:"maxTotalTasks,omitempty"`    // lifetime task count
```

These control task *volume*, not spending. A TaskSpawner with `maxTotalTasks: 100` could cost anywhere from $10 to $1500+ depending on task complexity. For teams managing agent budgets, this is the wrong abstraction.

### 3. No automatic circuit-breaker for runaway costs

When a TaskSpawner processes items that trigger unusually expensive agent runs (large codebases, complex prompts, long-running sessions), costs accumulate with no upper bound. The only manual intervention is `kubectl patch ... --type=merge -p '{"spec":{"suspend":true}}'`. There is no built-in mechanism to self-limit.

### 4. The infrastructure for cost-aware control already exists

Everything needed to implement cost budgets is already in place:

- **Cost capture**: `kelos-capture` extracts `cost-usd` from agent output (`internal/capture/usage.go:57-58`)
- **Result propagation**: `output_parser.go` stores results in `Task.Status.Results`
- **Prometheus metrics**: Per-spawner cost counters with namespace/type/spawner/model labels
- **Condition precedent**: `TaskBudgetExhausted` condition already exists for `maxTotalTasks` (`cmd/kelos-spawner/main.go:444-459`)
- **Self-suspension pattern**: Spawner checks `Spec.Suspend` at cycle start (`cmd/kelos-spawner/main.go:243`)

### Real-world scenarios

| Scenario | Impact without budgetPolicy |
|----------|----------------------------|
| Team allocates $500/month for agent work, runs 3 TaskSpawners | No way to split or enforce budget per spawner — manual monitoring only |
| Agent hits a complex issue, runs for 40+ minutes costing $20 | No per-task cost ceiling to flag anomalies |
| Cron-based spawner runs nightly scans, cost creeps from $2→$12/run over time | No alert or suspension until someone checks Grafana |
| PR-triggered spawner gets 50 PRs merged in one day (release day) | `maxConcurrency` throttles parallelism but total daily spend is unbounded |

### Spec addition to `TaskSpawnerSpec`

```go
// BudgetPolicy configures cost-aware governance for the TaskSpawner.
// When configured, the spawner tracks cumulative cost from completed
// tasks and enforces limits. Cost data comes from the "cost-usd"
// key in Task.Status.Results, which is populated by kelos-capture.
type BudgetPolicy struct {
    // MaxCostPerTask flags tasks that exceed this cost threshold in USD.
    // This is advisory (agents cannot be cheaply interrupted mid-run):
    // the spawner logs a warning and sets a Condition, but does not
    // terminate running tasks. Used for anomaly detection.
    // Example: "5.00"
    // +optional
    // +kubebuilder:validation:Pattern=`^\d+(\.\d{1,2})?$`
    MaxCostPerTask string `json:"maxCostPerTask,omitempty"`

    // MaxTotalCostUSD is the lifetime cost ceiling for this spawner.
    // Once cumulative cost reaches this value, the spawner auto-suspends.
    // Example: "500.00"
    // +optional
    // +kubebuilder:validation:Pattern=`^\d+(\.\d{1,2})?$`
    MaxTotalCostUSD string `json:"maxTotalCostUSD,omitempty"`

    // PeriodBudget sets a recurring cost limit that resets on a schedule.
    // +optional
    PeriodBudget *PeriodBudget `json:"periodBudget,omitempty"`
}

// PeriodBudget defines a recurring cost ceiling.
type PeriodBudget struct {
    // Period is the budget reset interval.
    // +kubebuilder:validation:Enum=daily;weekly;monthly
    Period string `json:"period"`

    // MaxCostUSD is the maximum allowed cost per period.
    // +kubebuilder:validation:Pattern=`^\d+(\.\d{1,2})?$`
    MaxCostUSD string `json:"maxCostUSD"`

    // Action defines what happens when the period budget is exceeded.
    // "suspend" auto-suspends the spawner until the period resets.
    // "skip" stops creating new tasks but continues polling (resumes
    //  when the period resets without manual intervention).
    // +kubebuilder:validation:Enum=suspend;skip
    // +kubebuilder:default=skip
    Action string `json:"action,omitempty"`
}
```

### Status additions to `TaskSpawnerStatus`

```go
// BudgetStatus tracks cost accumulation for budgetPolicy enforcement.
type BudgetStatus struct {
    // TotalCostUSD is the cumulative cost of all tasks created by this spawner.
    // +optional
    TotalCostUSD string `json:"totalCostUSD,omitempty"`

    // CurrentPeriodCostUSD is the cost accumulated in the current budget period.
    // +optional
    CurrentPeriodCostUSD string `json:"currentPeriodCostUSD,omitempty"`

    // CurrentPeriodStart is when the current budget period began.
    // +optional
    CurrentPeriodStart *metav1.Time `json:"currentPeriodStart,omitempty"`

    // LastHighCostTask is the name of the most recent task that exceeded
    // maxCostPerTask. Empty when no task has exceeded the threshold.
    // +optional
    LastHighCostTask string `json:"lastHighCostTask,omitempty"`
}
```

### New Condition type

Following the existing `TaskBudgetExhausted` pattern at `cmd/kelos-spawner/main.go:444`:

```
Type:    "CostBudgetExceeded"
Status:  True/False
Reason:  "PeriodLimitReached" | "LifetimeLimitReached" | "BudgetAvailable"
Message: "Period cost ($145.20) has reached daily limit ($150.00)"
```

### New printer column

```go
// +kubebuilder:printcolumn:name="Cost",type=string,JSONPath=`.status.budgetStatus.totalCostUSD`,priority=1
```

### 1. Daily budget for a PR review spawner

```yaml
kind: TaskSpawner
metadata:
  name: pr-reviewer
spec:
  maxConcurrency: 3
  when:
    githubPullRequests:
      labels: ["needs-review"]
  taskTemplate:
    type: claude-code
    # ...
  budgetPolicy:
    maxCostPerTask: "8.00"         # Flag runs > $8 as anomalies
    periodBudget:
      period: daily
      maxCostUSD: "100.00"         # $100/day ceiling
      action: skip                 # Stop creating tasks, resume next day
```

### 2. Monthly budget for a cron-based scanner

```yaml
kind: TaskSpawner
metadata:
  name: nightly-security-scan
spec:
  when:
    cron:
      schedule: "0 2 * * *"
  taskTemplate:
    type: claude-code
    # ...
  budgetPolicy:
    maxTotalCostUSD: "2000.00"     # Lifetime cap
    periodBudget:
      period: monthly
      maxCostUSD: "400.00"         # $400/month
      action: suspend              # Hard stop — requires manual unsuspend
```

### 3. Lightweight anomaly detection only

```yaml
kind: TaskSpawner
metadata:
  name: issue-worker
spec:
  when:
    githubIssues:
      labels: ["agent-ready"]
  taskTemplate:
    type: claude-code
    # ...
  budgetPolicy:
    maxCostPerTask: "10.00"        # Alert on expensive runs, no hard limit
```

### Cost accumulation (controller side)

`RecordCostTokenMetrics()` at `internal/controller/task_controller.go:888` already runs when a task completes. Extend it to also update the parent TaskSpawner's `Status.BudgetStatus.TotalCostUSD` and `Status.BudgetStatus.CurrentPeriodCostUSD`:

1. Read the `kelos.dev/taskspawner` label from the completed task
2. Fetch the parent TaskSpawner
3. Parse and add `cost-usd` to `BudgetStatus.TotalCostUSD` and `CurrentPeriodCostUSD`
4. If `maxCostPerTask` is set and exceeded, set `BudgetStatus.LastHighCostTask` and log a warning
5. Update TaskSpawner status

This mirrors how `TotalTasksCreated` is already incremented after task creation.

### Budget enforcement (spawner side)

The spawner's reconciliation loop at `cmd/kelos-spawner/main.go:340-357` already checks `maxConcurrency` and `maxTotalTasks` before creating tasks. Add a parallel check:

```go
// Enforce cost budget (checked alongside maxConcurrency/maxTotalTasks)
if budgetPolicy := ts.Spec.BudgetPolicy; budgetPolicy != nil {
    if exceeded, reason := isCostBudgetExceeded(budgetPolicy, ts.Status.BudgetStatus); exceeded {
        log.Info("Cost budget exceeded, skipping remaining items", "reason", reason)
        break
    }
}
```

### Period reset

On each spawner reconciliation cycle, check if the current period has elapsed:

```go
if periodBudget != nil && budgetStatus.CurrentPeriodStart != nil {
    if isPeriodExpired(periodBudget.Period, budgetStatus.CurrentPeriodStart.Time) {
        budgetStatus.CurrentPeriodCostUSD = "0.00"
        budgetStatus.CurrentPeriodStart = &metav1.Time{Time: currentPeriodStart(periodBudget.Period)}
    }
}
```

For `action: skip`, the spawner just skips task creation until the period resets. For `action: suspend`, the spawner sets `Spec.Suspend = true` — and the user must manually un-suspend after reviewing costs.

### Agent types without cost reporting

Not all agent types emit `cost-usd` (only `claude-code` currently does per `docs/agent-image-interface.md:94-95`). For spawners using other agent types, `budgetPolicy` can still enforce token-based limits via a future `maxTokensPerPeriod` field, or users can rely on `maxTotalTasks` alongside `budgetPolicy.maxCostPerTask` for anomaly detection on whatever data is available.

## Backward compatibility

- **Fully backward compatible**: `budgetPolicy` is optional; omitting it preserves current behavior
- **No migration needed**: TaskSpawners without `budgetPolicy` continue working unchanged
- **Graceful degradation**: If tasks don't report `cost-usd` (non-claude-code agents), cost tracking remains at zero — `periodBudget` effectively never triggers, and `maxCostPerTask` is a no-op. This is safe and expected.
- **CRD update**: Standard additive field — existing resources validate without changes

## Relationship to existing proposals

- **#889 (failurePolicy)**: Complements this proposal — failurePolicy prevents retrying failed items (saving cost), while budgetPolicy caps total spending across all items. Both are per-spawner governance fields.
- **#930 (Kueue integration)**: Kueue governs *compute resource* quotas (CPU/memory for agent pods). budgetPolicy governs *AI API cost* — a different dimension. They're complementary.
- **#907 (schedulingPolicy)**: schedulingPolicy controls *when* tasks run. budgetPolicy controls *whether* tasks run based on accumulated cost. Could interact: a schedulingPolicy blackout window would reduce period spend.

## Future extensions

A standalone `CostBudget` CRD could aggregate spending across multiple TaskSpawners, enabling team-level or namespace-level budgets. This per-spawner `budgetPolicy` is the right starting point — it requires no new controllers, follows the existing `maxTotalTasks` pattern, and covers the most common use case. A cross-spawner CRD can be layered on top later if needed.

knechtionscoding:
#### What type of PR is this?

/kind feature

Replaces the TaskSpawner-owned `budgetPolicy` with a Task-native design that enforces spending controls on any Task regardless of how it was created (CLI, YAML, webhook, or TaskSpawner).

The original approach tied cost governance to the TaskSpawner, which left direct Tasks outside the model. This redesign makes budget enforcement a first-class Task concern via three new API surfaces:

**New CRDs:**
- `TaskBudget` — namespace-scoped policy that gates Task admission using label selectors. Supports daily period accounting (timezone-aware) with cost and token limits. Blocked Tasks stay in `Waiting` phase with a `BudgetBlocked` condition until the period resets.
- `TaskRecord` — immutable terminal record created when a Task completes. Survives Task TTL deletion (no ownerReference) and serves as the durable accounting source for budget enforcement.

**New Task.status fields:**
- `status.usage` — structured `TaskUsage` (`costUSD` as `resource.Quantity`, `inputTokens`, `outputTokens`) populated from `status.results` on completion.
- `status.conditions` — typed conditions including `BudgetBlocked`.

**Admission semantics:**
- `TaskBudget.spec.taskSelector` matches Tasks by label (uses `metav1.LabelSelector`)
- Zero matching budgets → no gate; multiple matches → all must admit
- Budget enforcement sums `TaskRecord` usage within the current period
- Fail-closed: a misconfigured matching budget blocks admission and sets a `Degraded` condition on the TaskBudget
- Server-side label filtering on TaskRecord listing for performance

**TaskSpawner integration:**
- TaskSpawner remains a Task producer; attaches budget labels via `taskTemplate.metadata.labels`
- All old budget code removed: `BudgetPolicy`, `CostBudget`, `PeriodBudget`, `BudgetStatus`, `HighCostTaskRecord` types; spawner-side enforcement; `accumulateSpawnerCost`; `kelos_spawner_budget_*` Prometheus gauges; `TaskCountExhausted`/`TaskBudgetExhausted`/`CostBudgetExceeded` conditions

**Design decisions:**
- `resource.Quantity` for cost fields (type-safe, integrates with k8s tooling)
- MVP period: Daily only (Weekly/Monthly deferred until boundary semantics specified)
- `BudgetAction` enum with `Block` as sole value (future-proofs for `Warn` etc.)
- `NowFunc` on reconciler for deterministic testing
- TaskRecord name = Task UID (guaranteed unique, survives recreation)
- Operators responsible for TaskRecord retention/pruning (documented)

Fixes #977

- CRD manifests are regenerated via `make update` and included.
- The `TaskRecordSpec` is immutable after creation (XValidation: `self == oldSelf`).
- `sumPeriodUsage` pushes label filtering server-side via `client.MatchingLabelsSelector`; only time filtering is client-side.
- Invalid timezone on a matching budget fails closed (blocks task, sets Degraded condition) rather than silently passing.
- Tests cover: `usageFromResults`, `computePeriodBoundaries`, `checkLimitsExceeded`, `createTaskRecord` (including idempotency).
- Example manifest at `examples/14-task-budget/taskbudget.yaml`.

```release-note
Add Task-native budget enforcement via new TaskBudget and TaskRecord CRDs. TaskBudget uses label selectors to match Tasks and enforces daily cost/token limits based on durable TaskRecord accounting. Blocked Tasks wait in Waiting phase with a BudgetBlocked condition. Task.status.usage provides structured cost and token data. Removes TaskSpawner-specific budgetPolicy in favor of the namespace-scoped TaskBudget design.
```


knechtionscoding:
@gjkim42 I have run the api-reviewer and regular reviewer against the code until they were both happy for approval. So this should be ready for your review when you have time

knechtionscoding:
@gjkim42 bump

knechtionscoding:
@gjkim42 bump here

knechtionscoding:
@gjkim42 Bump for feedback.

gjkim42 (commented):
/kelos api-review

sorry for the late response.

knechtionscoding:
@gjkim42 resolved review comments

gjkim42 (commented):
can we add e2e tests for this feature? at least for taskrecord.
not sure how can we e2e test taskbudget...
Maybe... we have to make a fake agent image pretending to use a specific amount of tokens...? by printing it? since the budget detection is the stdout-based.

knechtionscoding:
@gjkim42 will look to add E2E tests

gjkim42 (commented):
/kelos review

knechtionscoding:
@gjkim42 fixed and re-ran reviewer. Came back clean for me.

knechtionscoding:
@gjkim42 🤞 that the reviewer is happy now. I've resolved outstanding comments

knechtionscoding:
@gjkim42 updated to resolve merge conflicts, should be ready for your review

knechtionscoding:
@gjkim42 bump here for another review

gjkim42:
## API design review

Verdict: REQUEST CHANGES

### P1 - Validate TaskBudget selectors at the API boundary
`TaskBudget.spec.taskSelector` embeds `metav1.LabelSelector`, and the generated schema accepts invalid selector operator/value combinations. The controller later calls `metav1.LabelSelectorAsSelector` and fail-closes invalid selectors by setting `Degraded` and blocking Task admission. That means an API-accepted malformed budget can affect Task scheduling in the namespace.

I originally considered requiring root `.spec`, but existing Kelos CRDs do not require top-level `.spec`, so I would not make that part of the blocker. The important fix is to reject malformed `spec.taskSelector` policies on create/update, e.g. with CEL validation or a validating webhook, then run `make update`.

### P2 - Store effective worker identity in TaskRecord
`TaskRecord.spec.type` / `spec.model` are documented as legacy `Task.spec.type` / `Task.spec.model`, and `createTaskRecord` writes those fields directly. For current preferred-shape Tasks using `spec.worker.type` / `spec.worker.model`, records will omit the effective type/model. Either define these fields as effective agent type/model and populate them via existing `resolveTaskType` / `resolveTaskModel`, or remove them if they are not part of the accounting API contract.

internal/controller/task_controller.go
```diff
@@ -170,6 +184,14 @@ func (r *TaskReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 			}
 		}
 
+		admitted, result, err := r.checkBudgetAdmission(ctx, &task)
```
gjkim42, line 191: [P1][both] This only runs budget admission in the Job creation path. `TaskReconciler` returns before this point for `spec.workerPoolRef` tasks, and `WorkerPoolReconciler` completes those tasks without setting `Usage` or creating `TaskRecord`s. Matching worker-pool Tasks are never blocked and their usage never contributes to future budget checks, even though TaskBudget is documented as applying to Tasks generally.

api/v1alpha2/taskbudget_types.go
```diff
@@ -0,0 +1,121 @@
...
+
+// TaskBudget defines observed-spend admission limits for Tasks.
+// When a Task's labels match a TaskBudget's taskSelector and the budget
+// is exceeded, the Task stays in Waiting phase with a BudgetBlocked
+// condition until the period resets.
+type TaskBudget struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   TaskBudgetSpec   `json:"spec,omitempty"`
```
gjkim42: [P1][codex] `spec` is optional at the top level (`omitempty`), and the generated CRD only requires `period`/`taskSelector` inside `spec` when `spec` exists. A spec-less TaskBudget is accepted, decodes to an empty selector plus empty period type, matches all Tasks, and then fail-closes admission for the namespace. Please require `spec` or reject/ignore spec-less budgets before admission.

internal/controller/task_budget_admission.go
```diff
@@ -0,0 +1,354 @@
...
+		task.Status.Message = fmt.Sprintf("Budget %q exceeded: %s", budgetName, reason)
+		meta.SetStatusCondition(&task.Status.Conditions, metav1.Condition{
+			Type:               "BudgetBlocked",
+			Status:             metav1.ConditionTrue,
+			Reason:             "BudgetExceeded",
+			Message:            fmt.Sprintf("Blocked by TaskBudget %q: %s", budgetName, reason),
+			ObservedGeneration: task.Generation,
+			LastTransitionTime: metav1.Now(),
+		})
+		return r.Status().Update(ctx, task)
```
gjkim42: [P1][codex] This writes Task status every time a matching budget remains exceeded, even when the phase, message, and `BudgetBlocked` condition are already unchanged. Because Tasks are watched, that write can immediately trigger another reconcile and another write instead of waiting for `RequeueAfter`. Please add a no-op check before updating status.

internal/controller/task_controller.go
```diff
@@ -1066,6 +1090,16 @@ func (r *TaskReconciler) updateStatus(ctx context.Context, task *kelos.Task, job
 		RecordCostTokenMetrics(task, results)
 	}
 
+	// Create a TaskRecord for budget accounting when task completes with usage data,
+	// including the retry-output path where usage is populated after the initial transition.
+	var gcRequeue time.Duration
+	if (setCompletionTime || retryOutputs) && task.Status.Usage != nil {
+		if err := r.createTaskRecord(ctx, task); err != nil {
```
gjkim42, line 1107: [P1][google] The Task status update has already committed by the time `createTaskRecord` runs. If this create fails transiently, the next reconcile sees the Task terminal with results already set and returns before this block, so record creation is never retried and budget usage is permanently undercounted.

internal/controller/task_budget_admission.go
```diff
@@ -0,0 +1,354 @@
...
+			}
+			if requeueAfter <= 0 {
+				requeueAfter = time.Second
+			}
+
+			return false, ctrl.Result{RequeueAfter: requeueAfter}, nil
+		}
+
+		// Best-effort update TaskBudget status even when not exceeded
+		r.updateBudgetStatus(ctx, budget, periodStart, periodEnd, used)
```
gjkim42: [P2][google] `TaskBudget.status.used` is only refreshed during admission. After an admitted Task completes and creates a TaskRecord, no matching budgets are updated or enqueued, so status can show pre-completion usage until another Task happens to trigger admission. That makes the new status contract stale for operators.

internal/controller/task_controller.go
```diff
@@ -1075,6 +1109,11 @@ func (r *TaskReconciler) updateStatus(ctx context.Context, task *kelos.Task, job
 		return ctrl.Result{RequeueAfter: outputRetryInterval}, nil
 	}
 
+	// Requeue to clean up TaskRecords with pending TTL expiry
```
gjkim42: [P2][codex] Scheduling TaskRecord GC from this Task reconcile is not sufficient. When the timer fires, a still-existing Job can make `updateStatus` return before GC, and if the source Task has been deleted the reconcile exits on NotFound. In quiet namespaces, records with TTL can persist indefinitely; GC needs an independent trigger/path.

internal/controller/task_controller.go
```diff
@@ -1389,6 +1428,126 @@ func RecordCostTokenMetrics(task *kelos.Task, results map[string]string) {
...
+			Name:      string(task.UID),
+			Namespace: task.Namespace,
+			Labels:    task.Labels,
+		},
+		Spec: kelos.TaskRecordSpec{
+			TaskRef: kelos.TaskReference{
+				Name: task.Name,
+				UID:  task.UID,
+			},
+			Type:                      task.Spec.Type,
```
gjkim42: [P2][codex] These fields come from deprecated top-level `task.Spec.Type` and `task.Spec.Model`. Tasks using current `spec.worker.type` / `spec.worker.model` will produce records missing the effective type/model. Please use `resolveTaskType` and `resolveTaskModel` here.

cmd/kelos-spawner/main.go
```diff
@@ -462,25 +462,6 @@ func runCycleWithSourceCore(ctx context.Context, cl client.Client, key types.Nam
 		ObservedGeneration: ts.Generation,
 	})
 
-	// Set TaskBudgetExhausted condition
```
gjkim42, line 465: [P2][google] Removing this condition drops the existing status signal for `maxTotalTasks` exhaustion while the limit enforcement remains. Spawners still stop creating tasks, but clients and tests can no longer observe that the limit was reached. Please keep the compatibility condition or provide an equivalent status signal.

gjkim42 (changes requested):
Review-all found blocking correctness issues in the TaskBudget/TaskRecord implementation. The strongest signal is the worker-pool bypass, which both independent reviewers flagged. Overall correctness: patch is incorrect until the P1s are addressed.

internal/controller/task_controller.go
knechtionscoding: Ah, yeah, I hadn't considered the integration between these two when I worked on this, will updat this.

api/v1alpha2/taskbudget_types.go
knechtionscoding: Good call. Will fix

gjkim42:
@knechtionscoding 
Sorry for the late response.
The reviews are mostly about the conflicts with the newly introduced WorkerPool-generated tasks.

cmd/kelos-spawner/main.go
knechtionscoding: fixed in: 945f58c9b1161ed941eb359c5800bce3384b5c5a

internal/controller/task_controller.go
knechtionscoding: fixed in: 945f58c9b1161ed941eb359c5800bce3384b5c5a

internal/controller/task_controller.go
knechtionscoding: fixed in: 945f58c9b1161ed941eb359c5800bce3384b5c5a

internal/controller/task_budget_admission.go
knechtionscoding: fixed in: 945f58c9b1161ed941eb359c5800bce3384b5c5a

internal/controller/task_controller.go
knechtionscoding: fixed in: 945f58c9b1161ed941eb359c5800bce3384b5c5a

internal/controller/task_budget_admission.go
knechtionscoding: fixed in: 945f58c9b1161ed941eb359c5800bce3384b5c5a

api/v1alpha2/taskbudget_types.go
knechtionscoding: fixed in: 945f58c9b1161ed941eb359c5800bce3384b5c5a

internal/controller/task_controller.go
knechtionscoding: fixed in: 945f58c9b1161ed941eb359c5800bce3384b5c5a

knechtionscoding:
Closing in favor of #1426

PR #1330, opened 2026-06-10
Add --deny-slack-connect-channels option to slack server

knechtionscoding:
#### What type of PR is this?

/kind feature

Adds a `--deny-slack-connect-channels` flag to the shared Slack server. When enabled, the bot denies service to Slack Connect (externally shared) channels — leaving on invite and failing closed when channel status cannot be verified.

We found an issue where someone accidentally invited and exposed internal docs and bot access to external channels. This flag guards against that by treating any channel that cannot be confirmed as internal as denied.

**How it works:**

1. Listens for the `member_joined_channel` event (filtered to the bot's own user ID).
2. Calls `conversations.info` to check `is_ext_shared` and `is_pending_ext_shared`.
3. If either is true, calls `conversations.leave` and skips posting the join message.
4. If the `conversations.info` call fails (missing scopes, rate limits, timeouts, channel lookup errors), the channel is denied as well (fail closed).
5. If the channel is confirmed internal, proceeds with normal behavior (posting the welcome message if configured).

The flag defaults to `false` so existing deployments are unaffected.

N/A

- Renamed from `--auto-leave-ext-channels` to `--deny-slack-connect-channels` to better reflect intent — this is a server-level policy, not just a leave action.
- The `shouldDenySlackConnect` method returns `true` whenever the channel is identified as external OR when the status cannot be determined (fail closed). This ensures we never post into a channel whose Slack Connect status is unknown.
- Uses the existing `enrichCallTimeout` (5s) for the `conversations.info` call and `postMessageTimeout` (5s) for the leave call.

```release-note
Add --deny-slack-connect-channels flag to kelos-slack-server to deny service to externally shared (Slack Connect) channels. The bot leaves on invite and fails closed when channel status cannot be verified.
```

### DIFF
--- a/internal/slack/handler.go
+++ b/internal/slack/handler.go
@@ -31,20 +31,23 @@ const (
 // matching TaskSpawners. It is the centralized equivalent of the per-TaskSpawner
 // SlackSource that previously ran in each spawner pod.
 type SlackHandler struct {
-	client      client.Client
-	log         logr.Logger
-	taskBuilder *taskbuilder.TaskBuilder
-	api         *goslack.Client
-	sm          *socketmode.Client
-	botUserID   string
-	botID       string
-	joinMessage string
-	cancel      context.CancelFunc
+	client                   client.Client
+	log                      logr.Logger
+	taskBuilder              *taskbuilder.TaskBuilder
+	api                      *goslack.Client
+	sm                       *socketmode.Client
+	botUserID                string
+	botID                    string
+	joinMessage              string
+	denySlackConnectChannels bool
+	cancel                   context.CancelFunc
 }
 
 // NewSlackHandler creates a new handler. Call Start to begin listening.
 // If joinMessageFile is non-empty, the bot posts its contents when added to a channel.
-func NewSlackHandler(ctx context.Context, cl client.Client, botToken, appToken, joinMessageFile string, log logr.Logger) (*SlackHandler, error) {
+// If denySlackConnectChannels is true, the bot denies service to Slack Connect
+// (externally shared) channels, leaving on invite and failing closed on lookup errors.
+func NewSlackHandler(ctx context.Context, cl client.Client, botToken, appToken, joinMessageFile string, denySlackConnectChannels bool, log logr.Logger) (*SlackHandler, error) {
 	api := goslack.New(botToken, goslack.OptionAppLevelToken(appToken))
 
 	authResp, err := api.AuthTestContext(ctx)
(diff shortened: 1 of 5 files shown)


knechtionscoding:
@gjkim42 small one, should be ready for review

knechtionscoding:
@gjkim42 bump

gjkim42 (changes requested):
Thanks for the small, focused PR.

I think the current behavior should fail closed when `--auto-leave-ext-channels` is enabled. Right now `shouldAutoLeave` returns `false` when `conversations.info` fails, so the caller falls through to the join-message path and can still post into a channel whose Slack Connect status could not be verified. That includes missing scopes, rate limits, timeouts, or channel lookup errors. For this safety feature, unknown should mean do not post or process, even if we do not leave the channel unless it is confirmed external.

Related design note: I think this flag may be better framed as a server-level policy like `--deny-slack-connect-channels`, not only `--auto-leave-ext-channels`. It should cover join messages, normal message routing, and slash commands. A separate server-level `--allowed-channels` / Helm `slackServer.allowedChannels` would also be useful if operators want the Slack server to operate only in known channel IDs; the existing TaskSpawner `channels` field should remain the per-spawner routing filter.

Suggested minimum fix for this PR: when the Slack Connect classification call fails, skip the join message and add a test that `chat.postMessage` is not called on `conversations.info` failure
(and consider flag renaming?)

knechtionscoding:
@gjkim42 updated

gjkim42 (approved):
looks great.
Thank you.

internal/slack/handler.go
```diff
@@ -143,15 +147,20 @@ func (h *SlackHandler) handleEventsAPI(ctx context.Context, evt socketmode.Event
...
-// event is silently ignored.
+// handleMemberJoinedChannel is called when any user joins a channel. When the
+// bot itself is added, it denies Slack Connect channels if configured (leaving
+// and skipping the join message). Otherwise it posts the configured welcome message.
 func (h *SlackHandler) handleMemberJoinedChannel(ctx context.Context, evt *slackevents.MemberJoinedChannelEvent) {
 	if evt.User != h.botUserID {
 		return
 	}
 
+	if h.denySlackConnectChannels {
```
gjkim42, line 158: [P1] Enforce Slack Connect denial before routing events

The new `denySlackConnectChannels` check only runs in `handleMemberJoinedChannel`. Normal messages and slash commands still reach `routeMessage` without checking the flag, so a bot already present in a Slack Connect channel, a channel that later becomes externally shared, an invite where `conversations.info` fails, or a failed `conversations.leave` can still create Tasks from that channel. Please apply the deny/fail-closed check before routing message and slash-command events, or persist denied channel state and block routing from those channels.

PR #1334, opened 2026-06-10
Fix skills.sh AgentConfig skills never reaching the agent

### LINKED ISSUE #1332: Skills.sh doesn't work
AgentConfig's skillsSh doesn't work. It downloads it in an init container but doesn't properly use it.

gjkim42:
#### What type of PR is this?

/kind bug

`AgentConfig.spec.skills` (skills.sh packages) downloaded packages in the `skills-install` init container but the agent never saw them, for two reasons:

1. `npx skills add -g` with `HOME` set to the plugin volume installs into hidden per-agent directories (`.claude/skills/` for claude-code, `.agents/skills/` for codex/cursor/opencode). Agent entrypoints discover plugins by globbing `$KELOS_PLUGIN_DIR/*/`, which never matches hidden directories, so the downloaded skills were never wired into the agent.
2. The Kelos agent type was passed as the skills.sh agent name (`-a <type>`). `gemini` is not a valid skills.sh agent name (the CLI calls it `gemini-cli`), so the init container exited 1 and Gemini tasks failed outright.

The install script now uses the `universal` agent target, which writes the canonical skill format to the fixed `$HOME/.agents/skills` path for every agent type, then relocates the result into `<mount>/skills-sh/skills/<skill>/` — the same `<plugin>/skills/<skill>/SKILL.md` layout inline plugins use, which all five agent entrypoints already consume. The script fails fast when nothing was installed, and `Build` rejects a user plugin named `skills-sh` while `spec.skills` is set so the relocation cannot collide mid-init with a confusing `mv` error.

The skills.sh e2e test now verifies the installed skill is actually visible to the agent: the prompt has the agent check for the installed `SKILL.md` and emit a marker assembled at runtime (`$(printf ...)`), so the log assertion cannot vacuously match the prompt echoed in the stream-json logs.

Fixes #1332

- Verified the skills.sh CLI behavior empirically: `-a claude-code` installs to `~/.claude/skills/`, `-a codex|cursor|opencode` to `~/.agents/skills/`, `-a gemini` exits 1 (invalid agent), `-a universal` to `~/.agents/skills/` for all types. Also simulated the generated init script locally end-to-end: skills land at `skills-sh/skills/<skill>/SKILL.md` and the entrypoint glob picks up `skills-sh/` as a plugin directory.
- Pre-existing limitation left out of scope: the codex/cursor/opencode entrypoints copy only `SKILL.md` and drop a skill's supporting files (`scripts/`, `references/`), which matters more now that full skills.sh packages flow through.

```release-note
Fixed AgentConfig `spec.skills` (skills.sh packages): installed skills are now exposed to the agent as a plugin named `skills-sh`, and Gemini tasks no longer fail during skill installation. The plugin name `skills-sh` is reserved in `spec.plugins[].name` when `spec.skills` is set.
```

---
## Summary by cubic
Fixes AgentConfig.spec.skills so `skills.sh` packages install and reach the agent via a `skills-sh` plugin; also resolves Gemini init failures. The e2e now uses a dedicated fixture and a deterministic install check to avoid false positives.

- **Bug Fixes**
  - Install with npx skills add -a universal -g, then move from $HOME/.agents/skills to <plugin>/skills/<skill>/SKILL.md so entrypoints load them via `skills-sh`.
  - Fail fast when nothing is installed; chown outputs; reject a user plugin named `skills-sh` when `spec.skills` is set.
  - E2E: switch to the `kelos-dev/e2e-skills:kelos-e2e` fixture, add an init container to verify `/kelos/plugin/skills-sh/skills/kelos-e2e/SKILL.md`, and assert on a unique marker in logs to prove the skill content reached the agent.
  - Docs: clarify `KELOS_PLUGIN_DIR` is set when plugins or skills are present, and that `spec.skills` appear under the `skills-sh` plugin.

- **Migration**
  - When using `AgentConfig.spec.skills`, do not name any plugin `skills-sh` in `spec.plugins[].name`.

<sup>Written for commit fd3889f66d698b7e8b6d8e8101806b84df2bf2a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1334?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/agent-image-interface.md
+++ b/docs/agent-image-interface.md
@@ -44,7 +44,7 @@ Kelos sets the following reserved environment variables on agent containers:
 | `KELOS_AGENT_TYPE` | The agent type (`claude-code`, `codex`, `gemini`, `opencode`, `cursor`) | Always |
 | `KELOS_BASE_BRANCH` | The base branch (workspace `ref`) for the task | When workspace has a non-empty `ref` |
 | `KELOS_AGENTS_MD` | User-level instructions from AgentConfig | When `agentConfigRef` is set and `agentsMD` is non-empty |
-| `KELOS_PLUGIN_DIR` | Path to plugin directory containing skills and agents | When `agentConfigRef` is set and `plugins` is non-empty |
+| `KELOS_PLUGIN_DIR` | Path to plugin directory containing skills and agents. Each subdirectory is one plugin in the `<plugin>/skills/<skill>/SKILL.md` layout; skills.sh packages from `spec.skills` appear under the `skills-sh` plugin | When `agentConfigRef` is set and `plugins` or `skills` is non-empty |
 | `KELOS_SETUP_COMMAND` | JSON-encoded exec-form array from `Workspace.spec.setupCommand`, executed by the entrypoint before the agent starts | When the workspace defines `setupCommand` |
 
 > The names listed in this table are reserved for Kelos behavior. When Kelos
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -218,7 +218,7 @@ GitHub Apps are preferred over PATs for production use because they offer fine-g
 | `spec.plugins[].skills[].content` | Skill content (markdown with frontmatter) | Yes (per skill) |
 | `spec.plugins[].agents[].name` | Agent name (becomes `agents/<name>.md`) | Yes (per agent) |
 | `spec.plugins[].agents[].content` | Agent content (markdown with frontmatter) | Yes (per agent) |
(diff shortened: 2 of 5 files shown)


gjkim42:
/kelos pick-up
Fix the CI

PR #1352, opened 2026-06-13
Reserve kelos-prefixed pod override volume names

### LINKED ISSUE #1343: Define reserved volume-name policy for podOverrides
## Problem

`Task.spec.podOverrides.volumes` is a user extension point, but Kelos also needs internal volume names for controller-managed pod wiring.

The current documented contract only reserves:

- `workspace`
- `kelos-plugin`

As Kelos adds more internal pod plumbing, reserving names one at a time can create controller-time compatibility breaks. A user manifest that previously used a volume name such as `kelos-plugin-src` could be accepted by the CRD and then fail later during Job construction after an upgrade.

## Proposal

Define an explicit policy for Kelos-managed volume names before adding more internal names.

Options to consider:

- Reserve all `kelos-*` volume names for Kelos-managed volumes.
- Keep only enumerated reserved names and make new internal volumes choose collision-free names dynamically.
- Use a hybrid policy: reserve `kelos-*` for new resources while preserving behavior for existing manifests where possible.

## API compatibility considerations

If Kelos reserves all `kelos-*` volume names, this is a tightening change for an existing free-form field. Existing manifests with user volumes such as `kelos-cache`, `kelos-tools`, or `kelos-config` would become invalid at controller time unless the migration is handled carefully.

If a reserved prefix is adopted, the policy should be reflected consistently in:

- Go API docs for `Task.spec.podOverrides.volumes`
- Generated CRDs
- Any TaskSpawner task template docs that embed `PodOverrides`
- Validation behavior, preferably admission-time rather than only Job-build-time
- Release notes, because this can reject previously accepted manifests

## Acceptance criteria

- Kelos has a documented policy for reserved volume names.
- Controller behavior and API documentation agree.
- Existing manifests either continue to work or the breaking behavior is explicitly documented with a migration path.
- Future internal volumes have a clear naming rule that does not create accidental user-volume collisions.

gjkim42:
#### What type of PR is this?

/kind feature

Reserves user-supplied Task pod override volume names matching `kelos-*` for Kelos-managed volumes, while continuing to reserve `workspace`. Adds API-server CEL validation, keeps controller validation in sync, regenerates CRDs, and documents the migration path for existing user volumes with `kelos-*` names.

Fixes #1343

`TaskSpawner.spec.taskTemplate.podOverrides.volumes` gets the same generated validation because it embeds `PodOverrides`.

```release-note
[ACTION REQUIRED] Kelos now reserves `workspace` and all `kelos-*` Task pod override volume names for controller-managed volumes. Rename user-defined `kelos-*` volumes and matching volume mounts before upgrading.
```


gjkim42:
/kelos pick-up
we cannot help with the tightened validation.
Let's update the release note of the PR by adding [ACTION REQUIRED] ...

PR #1416, opened 2026-07-02
Reuse GitHub status comments

gjkim42:
#### What type of PR is this?

/kind bug

Updates GitHub task status reporting to reuse an existing Kelos status comment for the same task instead of posting another comment when the stored comment ID annotation is missing. The reporter now searches existing issue/PR comments for the Kelos task-status marker, patches the latest matching comment, and only creates a comment when no match exists.

This keeps GitHub issue and PR timelines compact while preserving the existing initial-comment behavior for tasks that have not reported before.

N/A

Validated with:
- `make test`
- `make verify`

```release-note
GitHub task status reporting now updates an existing Kelos status comment for the task instead of creating additional lifecycle comments when possible.
```

---
## Summary by cubic
Updates GitHub status reporting to reuse a single Kelos status comment per task. When the stored comment ID is missing, we find the existing status comment and update it instead of creating duplicates.

- **Bug Fixes**
  - Added `FindTaskStatusComment` to scan issue/PR comments (paged) and find the latest Kelos status by header and task name; updates it when found, creates only if none exist.
  - Updated watcher to call this lookup when `AnnotationGitHubCommentID` is absent and persist the found ID for future runs.
  - Adjusted tests to cover GET→PATCH/POST flows and added coverage for not-found cases.
  - Clarified docs to say we maintain one status comment per task.

<sup>Written for commit 266997b8903c584a609f1539b6fa0f2d08759299. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1416?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/reporting/github.go
+++ b/internal/reporting/github.go
@@ -8,9 +8,14 @@ import (
 	"io"
 	"net/http"
 	"strconv"
+	"strings"
 )
 
-const defaultBaseURL = "https://api.github.com"
+const (
+	defaultBaseURL            = "https://api.github.com"
+	githubStatusCommentHeader = "🤖 **Kelos Task Status**"
+	commentsPageSize          = 100
+)
 
 // GitHubReporter posts and updates issue/PR comments on GitHub.
 // TokenFunc, when set, is called on every API request to resolve the current
@@ -45,7 +50,8 @@ type createCommentRequest struct {
 }
 
 type commentResponse struct {
-	ID int64 `json:"id"`
+	ID   int64  `json:"id"`
+	Body string `json:"body"`
 }
 
 // CreateComment creates a comment on a GitHub issue or pull request and returns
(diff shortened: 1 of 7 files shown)


internal/reporting/github.go
```diff
@@ -8,9 +8,14 @@ import (
 	"io"
 	"net/http"
 	"strconv"
+	"strings"
 )
 
-const defaultBaseURL = "https://api.github.com"
+const (
+	defaultBaseURL            = "https://api.github.com"
+	githubStatusCommentHeader = "🤖 **Kelos Task Status**"
```
knechtionscoding, line 16: Should this be configurable? So that if multiple agents run against the same PR they aren't all updating the same comment?

gjkim42:
closed in favor of #1612

PR #1434, opened 2026-07-07
Support GitHub App secrets for persistent WorkerPools

### LINKED ISSUE #1433: Support GitHub App token secrets for persistent WorkerPools
## Summary

Today the WorkerPool reconciler hard-rejects workspaces backed by a GitHub App
secret ([`workerpool_controller.go:178-187`](internal/reconciler/workerpool_controller.go#L178-L187)):

> `workspace %q uses a GitHub App secret which is not supported for persistent worker pools (use a PAT-style secret instead)`

We should allow a WorkerPool to use an App-backed workspace by minting and
refreshing a short-lived installation token, mirroring the pattern the ephemeral
Task path already implements.

## Background — why it's blocked today

Two independent problems combine into the block:

1. **The token expires (~1h) but a WorkerPool pod lives for hours/days.** A
   GitHub App secret holds `appID` / `installationID` / `privateKey` — not a
   usable token. Something must mint an installation token and re-mint it before
   expiry.
2. **The pool injects the token as a frozen env var.**
   [`workerpool_controller.go:491-493`](internal/reconciler/workerpool_controller.go#L491-L493)
   sets `GITHUB_TOKEN` via `valueFrom.secretKeyRef`. Secret-backed env vars are
   captured at pod start and never update, even if the secret changes. The
   worker's git credential helper falls back to `$GITHUB_TOKEN` because — unlike
   the Task path — the pool never sets `KELOS_GITHUB_TOKEN_FILE`. The runner
   freezes it further: [`runner.go:280`](internal/runner/runner.go#L280) passes
   `os.Environ()` (the pod-start `GITHUB_TOKEN`) to every task's agent
   subprocess.

So even a refreshed secret would not be picked up. Both halves need fixing:
(A) keep a valid token in a secret, and (B) make the pod consume the current
token rather than a frozen copy.

## What already exists (reuse, don't reinvent)

The ephemeral Task path already solves the hard parts:

- **Mint + refresh loop:** `resolveGitHubAppToken` mints a derived
  `<name>-github-token` secret annotated with source-secret name + expiry;
  `refreshGitHubAppToken` re-mints before expiry via a requeue keyed on
  `tokenExpiresAtAnnotation`.
- **Token provider:** `TokenProvider` with a 5-minute expiry margin
  (`tokenExpiryMargin`).
- **File-based consumption that survives refresh:** the ephemeral path sets
  `KELOS_GITHUB_TOKEN_FILE` ([`job_builder.go:461`](internal/reconciler/job_builder.go#L461))
  to a mounted secret-volume file; the shared git credential helper
  ([`job_builder.go:1042`](internal/reconciler/job_builder.go#L1042)) reads that
  file on each call, so kubelet's secret-volume sync propagates a refreshed
  token into a running pod.
- **Pool scaffolding stub:** [`workerpool_controller.go:515-519`](internal/reconciler/workerpool_controller.go#L515-L519)
  already sets `KELOS_TOKEN_SECRET` "so the worker runner reads this secret
  name" — but the runner never implements it.

### 1. Controller — replace the rejection with mint + refresh (`workerpool_controller.go`)

- Delete the guard at lines 178-187. When the workspace secret `IsGitHubApp`,
  mint a derived `<pool>-github-token` secret (owner-referenced to the pool) and
  point the StatefulSet at that secret instead of the raw App secret.
- Add a `RequeueAfter` return driven by the expiry annotation — same logic as
  `refreshGitHubAppToken`, refreshing at `expiresAt - tokenRefreshMargin`. The
  pool reconciler already returns `ctrl.Result{RequeueAfter: ...}` in several
  places.
- Refactor the mint/refresh bodies out of `TaskReconciler` into a shared helper
  (e.g. in `internal/githubapp` or a small `internal/tokensecret` package)
  taking `(client, namespace, ownerRef, appSecret, apiBaseURL)` so both Task and
  WorkerPool call one implementation.
- Confirm `WorkerPoolReconciler` has a `TokenClient` wired in
  `cmd/kelos-controller/main.go` (the Task reconciler does; the pool one likely
  needs it added).

### 2. Pod consumption — mount as a file, not (only) env (`buildStatefulSet`, `workerpool_controller.go:400`)

- Mount the derived token secret as a volume and set `KELOS_GITHUB_TOKEN_FILE`
  to the mounted path — mirroring the ephemeral Task path. This lets the
  existing `gitCredentialHelper()` read the fresh token from the file, so git
  operations survive a refresh mid-pod.
- Keep `GITHUB_TOKEN` / `GH_TOKEN` env for tools that read env directly, but
  treat it as only a startup value; the file is the source of truth.

### 3. Worker-runner — inject the current token per task (`taskAgentEnv`, `runner.go:285`)

- Before each task, read `KELOS_GITHUB_TOKEN_FILE` and set `GITHUB_TOKEN` /
  `GH_TOKEN` / `GH_ENTERPRISE_TOKEN` in the agent subprocess env from the file
  contents (overriding the stale `os.Environ()` values). Per-task freshness for
  env-reading tools, no new RBAC (it's a file read).
- Optional hardening for a single task running longer than the refresh margin:
  the credential helper already re-reads the file per git call, so git is
  covered; only long-lived env-cached clients would be stale — an acceptable v1
  limitation to document.

## Edge cases & trade-offs

- **Kubelet sync lag:** secret-volume updates propagate in up to ~1 min; the
  5-min `tokenRefreshMargin` covers this comfortably.
- **Controller downtime:** if the controller is down past expiry, the pool's
  token goes stale and tasks fail auth until it recovers — same failure profile
  as ephemeral Tasks today.
- **Enterprise base URL:** reuse the `parseGitHubRepo` → `gitHubAPIBaseURL`
  resolution already in `refreshGitHubAppToken` so GHE pools mint against the
  right host.
- **Why not have the runner mint tokens directly?** That would push the App
  private key into every worker pod and require GitHub API egress + RBAC from
  workers. Keeping minting in the controller and shipping only short-lived
  tokens is the safer boundary and matches the ephemeral design.

## Testing

**Unit:**
- Shared mint/refresh helper: table-test expiry math, in-place update,
  annotation round-trip.
- `buildStatefulSet`: assert the file volume + `KELOS_GITHUB_TOKEN_FILE` are
  present for App secrets.
- `taskAgentEnv`: overrides token env from a temp file.
- Replace the existing rejection test at `workerpool_controller_test.go:928`
  with a "mints derived secret" assertion.

**e2e:** a WorkerPool bound to an App-secret workspace runs a task that does an
authenticated git op. Hard to run locally per CLAUDE.md — validate via PR CI.

## Suggested sequencing

1. Extract the shared token-secret mint/refresh helper from `TaskReconciler`
   (lowest-risk, highest-leverage; no behavior change).
2. Wire `TokenClient` into `WorkerPoolReconciler` if missing.
3. Replace the pool rejection with mint + requeue-based refresh.
4. Add the file volume + `KELOS_GITHUB_TOKEN_FILE` to `buildStatefulSet`.
5. Update `taskAgentEnv` to read the token file per task.
6. Swap the rejection test for mint/refresh coverage; add `buildStatefulSet` +
   runner tests.

## Size estimate

Medium change, concentrated in `workerpool_controller.go` + one refactor of the
Task token helpers + a small runner change.

## Open questions

- Where should the shared helper live — `internal/githubapp` vs a new
  `internal/tokensecret`?
- Do any env-only GitHub clients in the agent images need sub-task refresh, or is
  per-task + git-credential-helper freshness sufficient for v1?
- Should the derived pool token secret be per-pool or per-pod? (Per-pool is
  simpler and matches how the StatefulSet references a single secret.)

knechtionscoding:
#### What type of PR is this?

/kind feature

Adds GitHub App–backed workspace support to persistent WorkerPools, which
previously hard-rejected App secrets with an error and required a PAT-style
secret instead.

The blocker was that a GitHub App installation token expires (~1h) while a
WorkerPool pod lives for hours or days, and the pod consumed the token as a
frozen, secret-backed env var captured at pod start. This PR fixes both halves:

- **Keep a valid token in a Secret.** The WorkerPool reconciler now resolves an
  App-backed workspace secret into a derived `<pool-name>-github-token` Secret
  holding a short-lived installation token, minting it on first reconcile and
  re-minting it before expiry. The minting/refresh logic is extracted from the
  Task reconciler into a shared `github_token_secret.go` helper so both
  reconcilers share one implementation.
- **Make the pod consume the current token.** The derived Secret is mounted as a
  file (`/kelos/github-token/GITHUB_TOKEN`), which the kubelet keeps in sync. The
  git credential helper already re-reads that file per git call; the worker
  runner now also refreshes `GITHUB_TOKEN` / `GH_TOKEN` / `GH_ENTERPRISE_TOKEN`
  from the file for each task's agent subprocess, so env-reading tools pick up
  refreshes without a pod restart.

The next token refresh is scheduled alongside the StatefulSet requeue, and the
WorkerPool reconciler's RBAC is widened to `create;update` on secrets.

Fixes #1433

- The token-minting/refresh code paths for Tasks and WorkerPools are now unified
  in `internal/controller/github_token_secret.go`; the Task reconciler behavior
  is unchanged by the refactor.
- PAT-style workspaces are unaffected: they skip minting entirely and continue
  to use the secret as-is.

```release-note
WorkerPools now support GitHub App–backed workspaces. The controller mints a short-lived installation token into a `<pool-name>-github-token` Secret and re-mints it before expiry, mounting it as a file so long-lived worker pods keep a valid token without restarts.
```

---
## Summary by cubic
Enable GitHub App–backed workspaces for persistent WorkerPools by minting and auto-refreshing a short‑lived installation token and mounting it as a kubelet-synced file. Long‑lived worker pods keep a valid token without restarts; PAT workspaces are unchanged.

- New Features
  - WorkerPools now support GitHub App secrets via a derived `'<pool-name>-github-token'` Secret that is minted on first reconcile and refreshed before expiry; if a preexisting token Secret isn’t controlled by the pool we refuse to use/overwrite it, and we re‑mint if the workspace’s App secret changes.
  - Token is mounted as a file at `/kelos/github-token/GITHUB_TOKEN`; kubelet keeps it in sync, and the StatefulSet exposes `KELOS_GITHUB_TOKEN_FILE` so the worker runner re-reads it.
  - Worker runner reads `KELOS_GITHUB_TOKEN_FILE` per task and refreshes `GITHUB_TOKEN`/`GH_TOKEN`/`GH_ENTERPRISE_TOKEN` for env-reading tools.
  - Controller RBAC widened to `create;update` on Secrets; `TokenClient` injected to mint installation tokens; docs updated.

- **Refactors**
  - Unified token mint/refresh logic into `internal/controller/github_token_secret.go` and switched Task and WorkerPool reconcilers to use it (Task behavior unchanged); per-call `TokenClient` selection avoids races across different GitHub/GHE hosts.

<sup>Written for commit 9a2a9d35b8833dc8471f7809436b1d96ed52f687. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1434?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cmd/kelos-controller/main.go
+++ b/cmd/kelos-controller/main.go
@@ -276,6 +276,7 @@ func main() {
 		Scheme:                      mgr.GetScheme(),
 		Recorder:                    mgr.GetEventRecorderFor("kelos-controller"),
 		Clientset:                   clientset,
+		TokenClient:                 githubapp.NewTokenClient(),
 		WorkerRunnerImage:           workerRunnerImage,
 		WorkerRunnerImagePullPolicy: corev1.PullPolicy(workerRunnerImagePullPolicy),
 		ClaudeCodeImage:             claudeCodeImage,
--- a/docs/agent-image-interface.md
+++ b/docs/agent-image-interface.md
@@ -124,15 +124,18 @@ failures when tailing pod logs.
 ### 8. GitHub token freshness
 
 For GitHub App-backed workspaces, the installation token has a ~1h TTL but
-the controller re-mints it in place while the Job is running. The
+the controller re-mints it in place while the workload is running. This
+applies to both per-Task Jobs and long-lived WorkerPool StatefulSets. The
 `GITHUB_TOKEN`, `GH_TOKEN`, and `GH_ENTERPRISE_TOKEN` env vars are still
 set for compatibility, but their values are frozen at pod start and will
-expire mid-run for long-running tasks.
+expire mid-run for long-running tasks and for pooled workers.
 
 **Custom agent images should read `$KELOS_GITHUB_TOKEN_FILE` on each
 GitHub call** instead of capturing the env var once. The file is mounted
-from the per-task token Secret; the kubelet auto-syncs its contents
-(~60s latency) so each read returns the current token.
+from the derived token Secret — `<task-name>-github-token` for per-Task
+Jobs and `<pool-name>-github-token` for WorkerPools — and the kubelet
+auto-syncs its contents (~60s latency) so each read returns the current
+token.
 
 Two concrete recommendations:
 
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -191,6 +191,8 @@ For each labeled Secret with a non-empty `CODEX_AUTH_JSON` key, the controller m
 
 A WorkerPool manages a fleet of persistent worker pods backed by a StatefulSet. Tasks reference a WorkerPool via `spec.workerPoolRef` to execute on pre-warmed infrastructure instead of creating per-task Jobs.
 
+A WorkerPool's Workspace may use either a PAT-style or a GitHub App secret. For App secrets the controller mints a short-lived installation token into a `<pool-name>-github-token` Secret and re-mints it before expiry (see [Workspace authentication](#workspace-authentication)), so long-lived worker pods keep a valid token without restarts.
+
 | Field | Description | Required |
 |-------|-------------|----------|
 | `spec.worker.type` | Agent type | Yes |
@@ -278,7 +280,7 @@ kubectl create secret generic github-app-creds \
 
 GitHub Apps are preferred over PATs for production use because they offer fine-grained permissions, higher rate limits, no dependency on a specific user account, and automatically expiring tokens.
 
-The installation token is minted to a per-task Secret (`<task-name>-github-token`) at admission and is mounted into the agent pod as a file at `/kelos/github-token/GITHUB_TOKEN`. While the task is running, the controller re-mints the token before it expires and updates the Secret in place. The kubelet syncs the file contents automatically, and the agent image's git credential helper and `gh` wrapper read the file on each invocation, so tasks that run longer than the ~1h installation-token TTL keep working without a pod restart.
+The installation token is minted to a derived Secret and mounted into the agent pod as a file at `/kelos/github-token/GITHUB_TOKEN`. For per-Task Jobs the Secret is `<task-name>-github-token` (minted at admission); for WorkerPools it is `<pool-name>-github-token` (minted on first reconcile). In both cases the controller re-mints the token before it expires and updates the Secret in place. The kubelet syncs the file contents automatically, and the agent image's git credential helper and `gh` wrapper read the file on each invocation, so runs that last longer than the ~1h installation-token TTL keep working without a pod restart. This makes GitHub App secrets usable for persistent WorkerPools, not just ephemeral Task pods.
 
 ## AgentConfig
 
--- a/internal/controller/github_token_secret.go
+++ b/internal/controller/github_token_secret.go
@@ -1,0 +1,209 @@
+/*
+Copyright 2026 Kelos contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/kelos-dev/kelos/internal/githubapp"
+)
+
+// githubTokenSecretName returns the name of the derived Secret that holds the
+// short-lived GitHub App installation token minted for the resource named
+// ownerName (a Task or a WorkerPool).
+func githubTokenSecretName(ownerName string) string {
+	return ownerName + "-github-token"
+}
+
+// tokenClientForRepo returns a per-call TokenClient targeting the GitHub API
+// host derived from repoURL, falling back to base's BaseURL. A per-call client
+// avoids racing on the shared client's BaseURL across concurrent reconciles
+// that resolve to different GitHub Enterprise hosts.
+func tokenClientForRepo(base *githubapp.TokenClient, repoURL string) *githubapp.TokenClient {
+	tc := &githubapp.TokenClient{
+		BaseURL: base.BaseURL,
+		Client:  base.Client,
+	}
+	if repoURL != "" {
+		host, _, _ := parseGitHubRepo(repoURL)
+		if apiBaseURL := gitHubAPIBaseURL(host); apiBaseURL != "" {
+			tc.BaseURL = apiBaseURL
+		}
+	}
+	return tc
+}
+
+// mintGitHubAppTokenSecret generates a new installation token from the given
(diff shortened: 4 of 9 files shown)


knechtionscoding:
## Review Summary

**Verdict**: APPROVE (minor, non-blocking suggestions)
**Scope**: Enables GitHub App–backed workspaces for persistent WorkerPools by minting a short-lived installation token into a derived `<pool-name>-github-token` Secret, re-minting before expiry, and mounting it as a kubelet-synced file so long-lived worker pods keep a valid token without restarts. Token mint/refresh logic is extracted into a shared `github_token_secret.go` helper reused by both the Task and WorkerPool reconcilers.

## Findings Overview

| P0       | 0     | — | none |
| P1       | 0     | — | none |
| P2       | 0     | — | none |
| P3       | 2     | workerpool_controller.go:317 / _test.go:1116 | mint path discards returned expiry; refresh-requeue not asserted in tests |

## Findings

### Correctness / Robustness
- [P3] [internal/controller/workerpool_controller.go:317](internal/controller/workerpool_controller.go#L317) — On the first-mint path, `mintGitHubAppTokenSecret` returns the token expiry but it is discarded (`if _, err := mintGitHubAppTokenSecret(...)`). The next refresh time is then re-derived by `refreshGitHubAppTokenSecret`, which does a fresh `Get` of the just-created Secret. With the controller's cache-backed client that read-after-create can momentarily miss (`IsNotFound`), in which case `refreshGitHubAppTokenSecret` returns `next=0` and **no refresh is scheduled for that reconcile**. It self-heals on the following reconcile (the StatefulSet `Owns` watch fires after the StatefulSet is created, and the Secret is in cache by then), so this is not a hard bug — but it's an avoidable window. Consider using the expiry returned by `mintGitHubAppTokenSecret` to compute `next = time.Until(expiresAt - tokenRefreshMargin)` directly on the mint path instead of depending on an immediate re-read.

### Tests
- [P3] [internal/controller/workerpool_controller_test.go:1116](internal/controller/workerpool_controller_test.go#L1116) — `TestWorkerPoolReconciler_MintsGitHubAppTokenSecret` discards the reconcile `Result` (`_, err = r.Reconcile(...)`). The central promise of this PR is that the token is *kept fresh*, which for WorkerPools hinges on `refreshAfter` being merged into the returned `Result.RequeueAfter`. With the synchronous fake client the re-read succeeds, so asserting `result.RequeueAfter > 0` (and `< ~1h`) would cheaply cover the refresh-scheduling path that is otherwise untested here. A test for the PAT-unchanged path (App-less secret → workspace SecretRef left untouched, no derived Secret) would also be worthwhile.

## Suggestions (optional)
- [P3] The shared helper `refreshGitHubAppTokenSecret` returning a 4-tuple `(time.Duration, time.Time, bool, error)` is a little dense at call sites. Not worth changing, just noting the two callers each destructure it slightly differently.

## Key takeaways
- Clean, well-scoped change. The refactor into `github_token_secret.go` genuinely unifies the Task and WorkerPool paths without altering Task behavior, and the file-mount + per-task env refresh correctly addresses the frozen-env-var problem for long-lived pods.
- The `KELOS_TOKEN_SECRET` → `KELOS_GITHUB_TOKEN_FILE` switch aligns WorkerPools with the existing Task/`gh`-wrapper contract; no stale consumers of the old var remain.
- RBAC widening to `create;update` on secrets, `Optional: true` on the token volume, non-fatal refresh-failure handling, per-repo `TokenClient` (GHE host safety), and docs are all consistent with the implementation. No blocking issues.

gjkim42 (commented):
generally looks good.
Can you review the comments? (then squash the commits)

knechtionscoding:
@gjkim42 fixed comments and then squashed.

gjkim42 (approved):

PR #1440, opened 2026-07-09
Fall back to legacy /rest/api/2/search when Jira Data Center 404s on /search/jql

### LINKED ISSUE #1439: Jira source 404s on Data Center/Server: /rest/api/2/search/jql is Cloud-only
## Summary

The Jira TaskSpawner source (`internal/source/jira.go`) always calls
`/rest/api/2/search/jql`, which discovery fails against Jira Data
Center/Server instances (self-hosted) with a 404. That endpoint is
Cloud-only — Data Center/Server only exposes the classic offset-paginated
`/rest/api/2/search`.

## Reproduction

Configured a `TaskSpawner` with a `jira` source (`baseUrl` pointing to a
Data Center instance, PAT auth per the "omit JIRA_USER" bearer-token path).
The spawner logs:

```
discovering items: Jira API returned status 404: {"message":"HTTP 404 Not Found","status-code":404,"sub-code":-1}
```

Confirmed directly against the instance:

- `GET /rest/api/2/search/jql?jql=...` → 404
- `GET /rest/api/2/search?jql=...` → 200, returns issues normally

`GET /rest/api/2/serverInfo` also returns 200, confirming the instance and
token are otherwise reachable/valid — this is purely an endpoint
availability mismatch between Jira Cloud and Data Center/Server.

## Expected

The Jira source should work against Data Center/Server instances (the doc
comment on the `Jira` CRD type explicitly describes PAT/Bearer auth for
"Jira Data Center/Server"), not just Cloud.

## Fix

I have a fix + tests ready: on a 404 from `/rest/api/2/search/jql`, fall
back to `/rest/api/2/search` with offset-based pagination for the rest of
that discovery run. Non-404 errors (e.g. auth failures) are returned
as-is without attempting the fallback, so real failures aren't masked.
Opening a PR shortly.

bmh8993:
/kind bug

Fixes #1439

## Summary
- The Jira source only called `/rest/api/2/search/jql`, which is a
  Cloud-only endpoint. Jira Data Center/Server (self-hosted) 404s on it,
  so discovery never returns any issues against those instances even
  though the `Jira` CRD type's doc comment explicitly documents PAT/Bearer
  auth for "Jira Data Center/Server".
- On a 404 from that endpoint, the source now falls back to the classic
  offset-paginated `/rest/api/2/search` endpoint for the remainder of the
  discovery run. Non-404 errors (auth failures, etc.) are returned as-is
  without attempting the fallback, so real failures aren't masked as a
  compatibility issue.

## Verification
- Reproduced against a real Jira Data Center instance: `/search/jql` →
  404, classic `/search` → 200 with real issues. Confirmed the fix
  resolves discovery there.
- Added two tests: one exercising the fallback across an offset-paginated
  two-page response, one confirming a non-404 error (401) is returned
  directly without hitting the legacy endpoint.
- `go build ./...` and `go test ./internal/source/...` pass; existing
  Jira tests (which only mock `/search/jql`) are unaffected since the
  fallback only triggers on 404.

## Test plan
- [x] `go test ./internal/source/...`
- [x] `go build ./...`
- [x] Manually verified against a live Jira Data Center/Server instance

```release-note
Fixed Jira issue discovery against Data Center/Server instances, which previously returned no issues because the source only called the Jira Cloud-only `/rest/api/2/search/jql` endpoint.
```

---
## Summary by cubic
Restore Jira issue discovery on Jira Data Center/Server by falling back to `/rest/api/2/search` when the Cloud-only `/rest/api/2/search/jql` returns 404. Resume from the current offset to avoid duplicates; non-404 errors are returned as-is.

- **Bug Fixes**
  - Detect 404 on `/search/jql` and switch to legacy `/search` for the rest of the run; resume with `startAt` based on already-fetched issues.
  - Support legacy pagination via `startAt` and `total`; keep token pagination for Cloud.
  - Add `jiraAPIError` and `isJiraNotFound` to expose status codes and drive fallback.
  - Add tests for two-page legacy fallback, mid-run resume, and non-404 pass-through; use t.Errorf in HTTP handler goroutines.

<sup>Written for commit c8c73392c6bd6221f65ed15043b5a7d77f990060. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1440?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/source/jira.go
+++ b/internal/source/jira.go
@@ -3,6 +3,7 @@ package source
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -36,6 +37,22 @@ type jiraSearchResponse struct {
 	Issues        []jiraIssue `json:"issues"`
 	NextPageToken string      `json:"nextPageToken"`
 	IsLast        bool        `json:"isLast"`
+
+	// StartAt and Total are only populated by the legacy /rest/api/2/search
+	// endpoint, which paginates by offset instead of NextPageToken.
+	StartAt int `json:"startAt"`
+	Total   int `json:"total"`
+}
+
+// jiraAPIError carries the HTTP status code returned by the Jira API so
+// callers can distinguish a missing endpoint (404) from other failures.
+type jiraAPIError struct {
+	StatusCode int
+	Body       string
+}
+
+func (e *jiraAPIError) Error() string {
+	return fmt.Sprintf("Jira API returned status %d: %s", e.StatusCode, e.Body)
 }
 
 type jiraIssue struct {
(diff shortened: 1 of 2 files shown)


gjkim42:
Could you sign the cla?

gjkim42:
Please squash commits

bmh8993:
Squashed. PTAL

gjkim42:
CI failed. please fix it

gjkim42:
@bmh8993 please squash commits.

bmh8993:
Squashed into a single commit. PTAL @gjkim42

bmh8993:
@gjkim42 
The Fork E2E run failed on a single spec that's unrelated to this PR:
`Session provider [codex] recovers an empty conversation after Pod recreation` (`test/e2e/session_test.go:506`) — 87 passed / 1 failed.

It's a flaky model-output failure, not a regression. The test prompts codex with `Join these fragments without spaces and reply with only the result: EMPTY_, SESSION_, READY` and expects `EMPTY_SESSION_READY`, but codex returned `EMPTY_,SESSION_,READY` (kept the comma separators). The same "Pod recreation" scenario passed for the `claude-code` and `opencode` providers. This PR only touches the Jira Data Center search fallback, which shares no code path with the session provider tests.

Could you re-run the failed E2E job? Should pass on retry. (Separately, the prompt is ambiguous — the comma-delimited fragments make it unclear whether the commas should be stripped — so this spec may be worth hardening in a follow-up.)

PR #1455, opened 2026-07-11
Recover Sessions with StatefulSets

gjkim42:
#### What type of PR is this?

/kind feature

Makes a Session a recoverable remote-control surface for one provider conversation.

- backs each Session with a one-replica StatefulSet and a headless governing Service
- adds optional `Session.spec.volumeClaimTemplate`; configured Sessions persist the workspace and runtime journal in a StatefulSet volume claim, while omitted storage uses `emptyDir`
- restores persisted conversation events after Pod replacement and marks unfinished work as interrupted without replaying its prompt
- reconnects terminal and web clients to replacement Pods while preserving accepted-request semantics
- covers StatefulSet ownership, optional storage, journaling, request delivery, and recovery behavior in unit, integration, and end-to-end tests

N/A

The headless Service is only the StatefulSet governing Service. Terminal and web clients continue to use the shared Session control path rather than connecting through a per-Session Service.

Pod replacement preserves workspace and conversation state only when `spec.volumeClaimTemplate` is configured. A Session that omits it receives a fresh `emptyDir` after replacement.

Validated locally with `make update`, `make build`, `make test`, `make verify`, and `make test-integration` (123 controller/CLI integration specs and 12 install-integration specs). The full e2e suite runs in CI.

```release-note
Sessions now run as StatefulSets and can preserve their workspace and conversation events across Pod replacement when configured with `spec.volumeClaimTemplate`; terminal and web clients reconnect automatically.
```


knechtionscoding (commented):
@gjkim42 this seems like an exceptional amount of code/work to handle what could be handled by having session pods use a stateful set?

PR #1533, opened 2026-07-21
fix: Add nameTemplate to TaskSpawner for deterministic Task naming and dedup

### LINKED ISSUE #1155: Webhook handler creates duplicate tasks when multiple GitHub events fire for the same PR
## Description

The webhook handler creates duplicate Tasks when multiple GitHub webhook events arrive for the same PR within a short window. This happens because task names are derived from `sha256(deliveryID)` — each unique webhook delivery gets a different name, bypassing any dedup.

## Observed behavior

A single PR spawned multiple tasks (1 Running, 4 Waiting)

Tasks were created within 1 second at 20:01 UTC, likely from a label re-apply or PR state change that triggered multiple matching webhook filters (`labeled`, `opened`, `ready_for_review`).

## Root cause

In `internal/webhook/handler.go`, `createTask` generates the task name as:
```go
sum := sha256.Sum256([]byte(deliveryID))
shortHash := hex.EncodeToString(sum[:])[:12]
taskName := fmt.Sprintf("%s-%s-%s", spawner.Name, sanitizedEventType, shortHash)
```

Since each GitHub webhook delivery has a unique `X-GitHub-Delivery` header, every event creates a differently-named task. The in-memory `DeliveryCache` only deduplicates identical deliveries (same delivery ID), not different events for the same PR.

The polling-based spawner in `cmd/kelos-spawner/main.go` uses the PR number as the task name suffix (`<spawner>-<prNumber>`), which naturally deduplicates. The webhook handler lacks an equivalent check.

## Expected behavior

When a webhook event fires for a PR that already has an active (Pending/Running/Waiting) task from the same spawner, the handler should skip task creation.

knechtionscoding:
#### What type of PR is this?

/kind feature

Adds an optional `nameTemplate` field to `TaskSpawner.spec.taskTemplate` that
renders the spawned Task's name from a Go text/template.

This gives operators control over Task names and, more importantly, a way to
**deduplicate Tasks**. GitHub frequently sends several distinct webhook
deliveries for a single pull request within a second (e.g. an `opened` event
plus one `labeled` event per label). Because the webhook handler derives Task
names from `sha256(deliveryID)`, each of those deliveries has a unique
`X-GitHub-Delivery` header and therefore creates a separately named Task — so
one PR spawns multiple duplicate Tasks (issue #1155).

With a deterministic `nameTemplate` (e.g. `"{{.Number}}"`), every delivery for
the same PR renders to the same Task name. The first delivery creates the Task;
later ones collide on the name and are treated as deduplication, so exactly one
Task is created. This mirrors how the polling spawner already deduplicates by
naming Tasks `<spawner>-<item.ID>`.

Behavior details:

- `nameTemplate` is optional. When unset, the polling, webhook, and Slack paths
  keep their existing default naming, so this is fully backward compatible.
- The rendered value is lowercased and sanitized into a valid Kubernetes
  resource name (invalid characters replaced with `-`, each DNS label trimmed of
  `-` and empty labels dropped so values like `a..b`/`a.-b` become `a.b`), then
  truncated to 63 characters. The result is validated with Kubernetes' DNS-1123
  helper; a value that renders empty or invalid is rejected up front rather than
  failing later at Task creation.
- `.Context.NAME` (context sources) is excluded from name rendering on every
  source, so a Task's identity never depends on mutable external data; a
  `nameTemplate` that references `.Context` fails to render.
- Deduplication is **ownership-scoped** on every creation path. A rendered
  name's uniqueness scope is the whole namespace, so an `AlreadyExists` collision
  is treated as a benign duplicate only when the existing Task belongs to the
  same TaskSpawner (matched by controller owner-reference UID, falling back to
  the `kelos.dev/taskspawner` label for ownerless legacy Tasks). A collision with
  an unrelated Task surfaces an actionable error instead of silently dropping
  work.
- Failures surface instead of being swallowed: the webhook handler returns
  spawner errors so the delivery responds HTTP 500 and GitHub retries; the
  polling cycle fails when a name cannot be resolved (bad template, missing key,
  or empty result) rather than reporting a successful zero-Task discovery.
- The polling spawner computes its deduplication lookup key from the rendered
  name so it matches the name the Task is actually created with.
- Manual CLI runs (`kelos run --from-taskspawner`) intentionally ignore
  `nameTemplate`: they manage their own name (an explicit `--name` or a unique
  `-manual-` suffix), and the template's variables may not be available for a
  manual invocation.
- The field is added only to the latest API version (`v1alpha2`). `v1alpha1`
  remains served; because it has no such field, `nameTemplate` is preserved
  across a v1alpha1 read/modify/write round-trip via an internal conversion
  annotation (the capability is not added to v1alpha1).

Fixes #1155

- This is the explicit, opt-in fix discussed on the issue (the maintainer's
  proposed `taskName`/template approach). It does not add automatic, source-
  specific deduplication when `nameTemplate` is unset — operators enable the
  behavior by configuring the field.
- The live `self-development/` spawners are intentionally **not** changed.
  None of them is the duplicate-firing "babysitter" case from the issue, and
  the `kelos-pr-responder` is deliberately re-triggered by `/kelos pick-up`
  comments — giving it a deterministic name would make repeat pick-ups on the
  same PR silently skip until its TTL expired, which would be a regression.
  Enabling `nameTemplate` on real spawners is better done as a separate,
  deliberate config change.
- Multi-repo / repeatable-trigger guidance is documented: a rendered name must be
  namespace-unique, so scope a dedup spawner to one repository (or include a
  bounded identifier before an unbounded `{{.Repository}}`, given the 63-char
  cap), and reserve deterministic naming for burst events — repeatable
  `issue_comment`/`pull_request_review` commands want a fresh Task each time. The
  webhook example is split accordingly.
- Tests added/updated:
  - `internal/taskbuilder`: name rendering, sanitization incl. DNS-label
    normalization and DNS-1123 validity, truncation, empty/invalid rejection,
    `.Context` exclusion, and `TaskBelongsToSpawner` (owner-UID precedence).
  - `internal/webhook`: duplicate deliveries collapse to one Task; distinct PRs
    stay separate; an unrelated name collision returns HTTP 500 (retryable).
  - `cmd/kelos-spawner`: `nameTemplate` applied; dedup lookup uses the rendered
    name; resolution error fails the cycle; cross-spawner collision errors.
  - `internal/slack`: cross-spawner name collision errors.
  - `internal/conversion`: `nameTemplate` survives a v1alpha1 round-trip.
  - `internal/cli`: manual runs ignore `nameTemplate`.
  - `internal/helmchart`: updated the CRD doc-string occurrence count for the
    added field.
- Generated files (CRD YAMLs) were regenerated with `make update`;
  `make verify` and `make test` pass.

```release-note
TaskSpawner task templates now support a `nameTemplate` field that renders the spawned Task's name from a Go text/template. Configure a deterministic template (e.g. `nameTemplate: "{{.Number}}"`) to deduplicate Tasks so that multiple GitHub webhook deliveries for the same pull request reuse a single Task instead of creating duplicates.
```

---
## Summary by cubic
Adds an optional `nameTemplate` to `TaskSpawner.spec.taskTemplate` to render deterministic Task names and deduplicate repeated GitHub webhook deliveries, Slack messages, and polling/cron items so they reuse a single Task. Dedup is ownership-scoped, errors surface on status, and aborted cycles do not advance cron watermarks.

- **New Features**
  - `nameTemplate` renders the Task name; output is lowercased, DNS-1123–sanitized, and truncated to 63 chars (empty/invalid is rejected). `.Context.*` is not available; referencing it fails. Keep the identifying part within 63 chars and ensure names are unique across repos (include `{{.Repository}}` when needed).
  - Dedup semantics: polling computes its lookup key from the rendered name; webhooks pre-check by resolved name (before context fetch) and skip if the existing Task belongs to the same spawner; Slack does the same. Cross-spawner collisions return an error (webhooks respond HTTP 500). When `nameTemplate` is unset, default naming is unchanged. Manual CLI runs (`kelos run --from-taskspawner`) ignore `nameTemplate`; `--name` or the manual suffix is used.
  - Status and reliability: per-cycle creation errors (e.g., name resolution or cross-spawner collisions) set a `DiscoveryError` condition and do not count as successful runs; aborted cycles (e.g., unresolvable `nameTemplate`) do not advance `LastDiscoveryTime` for cron sources.
  - GitHub webhook templates for issue/PR events expose `.Number`, `.Title`, `.Body`, `.URL` (plus `.Branch` for PRs). Available in API v1alpha2; CRDs, `docs/reference.md`, examples, and the webhook/Slack handlers were updated. API v1alpha1 round-trips preserve `nameTemplate` via an internal annotation.

- **Migration**
  - No changes needed if `nameTemplate` is unset.
  - To deduplicate PR webhooks, set `taskTemplate.nameTemplate: "responder-{{.Repository}}-pr-{{.Number}}"` (or similar).

<sup>Written for commit c727c70863a13f7c7d847eb13b9b72db7ca924b9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1533?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/v1alpha2/taskspawner_types.go
+++ b/api/v1alpha2/taskspawner_types.go
@@ -901,7 +901,7 @@ type TaskTemplate struct {
 	// Available variables (all sources): {{.ID}}, {{.Title}}, {{.Kind}}
 	// GitHub issue/Jira sources: {{.Number}}, {{.Body}}, {{.URL}}, {{.Labels}}, {{.Comments}}
 	// GitHub pull request sources additionally expose: {{.Branch}}, {{.ReviewState}}, {{.ReviewComments}}
-	// GitHub webhook sources: {{.Event}}, {{.Action}}, {{.Sender}}, {{.Ref}}, {{.Repository}}, {{.Payload}} (full payload access)
+	// GitHub webhook sources: {{.Event}}, {{.Action}}, {{.Sender}}, {{.Ref}}, {{.Repository}}, {{.Payload}} (full payload access); issue and pull request events also expose {{.Number}}, {{.Title}}, {{.Body}}, {{.URL}} (plus {{.Branch}} for pull requests)
 	// Linear webhook sources: {{.Type}}, {{.Action}}, {{.State}}, {{.Labels}}, {{.IssueID}}, {{.Payload}}
 	// Cron sources: {{.Time}}, {{.Schedule}}
 	// When contextSources are configured: .Context.NAME for each source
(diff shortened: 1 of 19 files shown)


knechtionscoding:
@gjkim42 ready for your review

knechtionscoding:
@gjkim42 resolved comments from cubic and api-review

gjkim42:
Don't consider the field drop during the roundtrip.
/kelos api-review

gjkim42 (commented):
Look good can you add an e2e test for this?
(You can address bot's review findings if you think it's worthwhile)

knechtionscoding:
@gjkim42 addressed bot feedback where relevant and added an E2E test

gjkim42:
can you rebase on origin/main?

I fixed the e2e workflow failure.

knechtionscoding:
@gjkim42 done

PR #1562, opened 2026-07-24
Add idle delete policy for Sessions

### LINKED ISSUE #1543: Add support for an idle TTL for a session
If a user has started a session and then leaves it idle for a period of time we should automatically reap it. Related to allowing to scale to 0 but not the same.

knechtionscoding:
#### What type of PR is this?

/kind feature

Adds an optional idle lifecycle policy to the Session API so abandoned Sessions
are reclaimed automatically instead of retaining their Pod and workspace storage
indefinitely.

```yaml
spec:
  idlePolicy:
    deleteAfterSeconds: 86400
```

`spec.idlePolicy.deleteAfterSeconds` deletes a Session once it has been
continuously idle for the configured number of seconds. Deletion cascades to
the Session's owned StatefulSet, Pod, workspace PersistentVolumeClaim, Service,
and RBAC.

Behavior:
- Idleness is measured from the later of `status.lastActivityTime` and the
  Session creation time. Renewed activity resets the idle period.
- A Session is only considered idle while its `Active` condition is explicitly
  `False`. An active turn (`True`) or unreported activity (`Unknown`, e.g. while
  the Pod is being replaced) never counts as idle, so in-flight work is never
  interrupted.
- Before deleting a Session with a live Pod, the controller drains the runtime
  via a dedicated idle-drain handshake: it sets an idle-drain request, the
  runtime stops accepting new turns and reports `Drained` once no turn is in
  flight, and only then does the controller delete. This closes the window
  where the runtime has locally accepted a turn but not yet published it to the
  Session status (notably the `deleteAfterSeconds: 0` case). The handshake
  reuses the existing runtime-update drain `Request`/`Report` types on dedicated
  annotations.
- Idle expiry is evaluated only after the `Active` condition has been validated
  against the current Pod, and the deletion is additionally guarded by a UID and
  resource-version precondition. Expired idle Sessions whose workload is already
  gone are reaped directly, since no turn can run without a Pod.
- The remaining time is folded into the reconcile `RequeueAfter`, so an idle
  Session is deleted on time without depending on external events.
- `deleteAfterSeconds` is unset by default (never deleted); `0` makes a Session
  eligible for deletion as soon as it goes idle.

The settings are grouped under an `idlePolicy` object rather than a flat TTL
field. This makes the outcome explicit (deletion, not suspension), avoids TTL
terminology, and leaves room for a future scale-to-zero change to add an
idle-triggered `suspendAfterSeconds` alongside it — the automatic counterpart to
the existing manual `spec.suspend`. When both are eventually configured, both
would be measured from the same continuous idle period and validation would
require `suspendAfterSeconds <= deleteAfterSeconds`.

Fixes #1543

- Only the implemented deletion behavior is exposed. `suspendAfterSeconds` is
  intentionally left out per the "keep API surfaces minimal" convention and will
  land with the scale-to-zero work.
- `spec.idlePolicy` is added only to the latest API version (`v1alpha2`);
  `Session` has no `v1alpha1`, so there is no conversion to update.
- `spec.idlePolicy` is immutable after creation (CEL validation, matching the
  sibling spec fields), so a live patch cannot lower `deleteAfterSeconds` and
  make a Session immediately eligible for destructive deletion.
- The field is additive and optional, so existing Session manifests and
  in-cluster resources continue to apply unchanged. It also flows into
  SessionSpawner's embedded Session template via codegen.
- The Session controller RBAC role gains the `delete` verb on `sessions`.
- Generated files (deepcopy, session/sessionspawner CRDs, RBAC, install-crd)
  were regenerated with `make update`. `make verify` and `make test` pass.

```release-note
Sessions now support `spec.idlePolicy.deleteAfterSeconds`, which automatically deletes a Session once it has been continuously idle (no active turn and no reported activity) for the configured number of seconds. Deleting the Session also removes its workspace storage. The field is optional; omit it to disable automatic idle deletion.
```


gjkim42:
Looks great.

I am considering to introduce a ttl field for pausing(scale replicas to 0)

Both should be considered for field naming.

gjkim42:
Sorry for the latest comment(deleted)

What do you think about the new api field suggestion?

knechtionscoding:
@gjkim42 I like it. My only concern was deviating from convetion (othe rk8s objects expose TTLSeconds), but in this case because we have two separate behaviors I think it makes sense to group the. I'll update my PR

knechtionscoding:
@gjkim42 updated based on comments local /review and /api-review came back clean

knechtionscoding:
Resolved comments @gjkim42

gjkim42 (commented):
Please add an e2e test for this as well

@knechtionscoding

knechtionscoding:
@gjkim42 added E2E test

knechtionscoding:
@gjkim42 resolved review comments

gjkim42 (approved):
Lgtm, thanks!

PR #1565, opened 2026-07-24
feat: expose changed files as {{.ChangedFiles}} in webhook prompt templates

knechtionscoding:
#### What type of PR is this?

/kind feature

Kelos already fetches a pull request's changed-file list for `filePatterns` filtering (`fetchPRChangedFilesWithToken` in `internal/webhook/github_files.go`) and then discards it once matching is done. This PR exposes that list to prompt templates as `{{.ChangedFiles}}` via `ExtractGitHubWorkItem`, so spawners can pass the set of touched files into the task prompt without an extra runtime `gh` call.

Details:
- `ChangedFiles` is **always** present in the template var map so `{{.ChangedFiles}}` never trips the renderer's `missingkey=error`.
- For PR events it is only populated when a matching filter's `filePatterns` forced the changed-files fetch (fetching is gated on `filePatterns`); otherwise it renders as an empty list.
- Push events populate `{{.ChangedFiles}}` from the payload regardless.
- Iterate with `{{range .ChangedFiles}}`.

This resolves the "deferred to a follow-up" NOTE previously left on the `ChangedFiles` field. Scope is intentionally limited to `.ChangedFiles` only — `.Diff` would require fetching per-file patch content (larger payloads, truncation concerns) and is left as a future change, consistent with keeping API surfaces minimal.

N/A

`{{.ChangedFiles}}` for PR events depends on `filePatterns` being set on the matching filter, because that is the only thing that triggers the lazy fetch. This coupling is documented in `docs/reference.md`. Adding a `.Diff` variable or fetching files when the prompt references `{{.ChangedFiles}}` without `filePatterns` are possible follow-ups.

```release-note
TaskSpawner/SessionSpawner GitHub webhook prompt templates can now reference `{{.ChangedFiles}}`, the list of changed file paths. It is populated for push events, and for pull request events when the matching filter declares `filePatterns`.
```

### DIFF
--- a/internal/webhook/github_filter.go
+++ b/internal/webhook/github_filter.go
@@ -46,9 +46,9 @@ type GitHubEventData struct {
 	// ChangedFiles lists file paths modified by the event.
 	// For push events, populated from the payload. For PR events, lazily
 	// fetched from the GitHub API when a webhook filter uses FilePatterns.
-	// NOTE: intentionally not exposed in ExtractGitHubWorkItem template vars
-	// yet — the {{.ChangedFiles}} template variable is deferred to a follow-up
-	// to resolve API design questions (slice vs pre-joined string, fetch gating).
+	// Exposed to prompt templates as {{.ChangedFiles}} via ExtractGitHubWorkItem.
+	// Because PR fetching is gated on FilePatterns, {{.ChangedFiles}} is only
+	// populated for PR events when a filter's filePatterns forced the fetch.
 	ChangedFiles []string
 	// Tag is the tag name for create (ref_type=tag) and release events.
 	Tag string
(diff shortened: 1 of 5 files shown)


internal/webhook/github_filter.go
```diff
@@ -688,6 +688,11 @@ func ExtractGitHubWorkItem(eventData *GitHubEventData) map[string]interface{} {
 		"ID":    eventData.ID,
 		"Title": eventData.Title,
 		"Kind":  "webhook",
+		// Files is the list of changed file paths. It is always present (so
+		// {{.Files}} never trips missingkey=error) but is only populated when a
+		// filter's filePatterns forced a changed-files fetch; otherwise it is
+		// empty. See the ChangedFiles field for details.
+		"Files": eventData.ChangedFiles,
```
gjkim42: Is it better to name it ChangedFiles?

gjkim42 (commented):
Looks good, one nit for naming.

internal/webhook/github_filter.go
knechtionscoding: I'm fine with changedFiles

knechtionscoding:
That feels like a bug to fix now, so I am going to do so.

knechtionscoding:
@gjkim42 this should be ready

PR #1588, opened 2026-08-04
Add TaskScore for external scoring of agent results

### LINKED ISSUE #1586: Let external actors score the results of Kelos agents
## Problem

Kelos has no way to learn whether an agent's output was any good. We can see
that a Task succeeded, what it cost, and how many tokens it burned — but
"Succeeded" only means the process exited cleanly. It says nothing about whether
the answer was useful, whether the PR was worth reviewing, or whether a given
spawner and model combination is actually earning its budget.

The people who already know the answer are outside the cluster. A reviewer reads
the agent's Slack reply and reacts to it. A maintainer approves or rejects the
PR the agent opened. That judgement is being expressed today and thrown away.

We want to capture it, consistently across every spawner, and in a form that
scales past a handful of tasks per day.

## Why this is not just "add a field"

Reporting is currently outbound-only. When a Task finishes, the reporting
watcher posts a Slack thread reply or a GitHub comment and records provenance
about the *originating* item (`kelos.dev/source-*`, `slack-thread-ts`,
`github-comment-id`). Nothing lets an inbound event find its way back to a Task:

1. **The Slack result message timestamp is never persisted.** It lives only in
   an in-memory map keyed by Task UID in `SlackTaskReporter`, so it dies with
   the pod. A 👍 on the agent's answer cannot be resolved to a Task at all.
2. **There is no index from external artifact to Task.** Even the GitHub comment
   ID is stored as an annotation, which the API server cannot select on.
   Resolving an inbound reaction would mean listing and scanning every Task —
   fine at ten tasks a day, not at ten thousand.
3. **Tasks are TTL-deleted.** Scores arrive minutes or days later. The durable
   anchor has to be the TaskRecord, not the Task.

So the correlation path has to be built before any scoring feature can work.

## Design constraints

**Consistency across spawners is the main risk.** Per-trigger filtering has
already diverged badly — `github_filter.go` is ~840 lines with near-parallel
logic re-implemented in `linear_filter.go`, `generic_filter.go`, and
`slack/filter.go`. If scoring is added per-source it will drift the same way.
Normalization must happen once, at a shared boundary, and every source must map
its own vocabulary into the same small verdict enum. Sources keep the
untranslated signal so verdicts can be re-derived when a mapping turns out
wrong.

**Reception and outcome are different axes.** Conflating them makes the dataset
worthless:

- *Reception* — did a human like the answer (reaction, review, button). Cheap,
  dense, biased toward tone and formatting.
- *Outcome* — did the work land and stick (PR merged, no revert within N days,
  CI green on the agent's commit). Sparse, delayed, actually correlated with
  value.

A PR merged after five rounds of corrections is a bad first-try score and a good
outcome score. These need to stay separate signals, aggregated separately.

**Scores must be append-only and retractable.** Multiple actors can score the
same result. Someone can react and then un-react. Read-modify-write on a single
counter field would produce conflict storms at scale — the same reason
TaskRecord itself is immutable.

## Proposed shape

- A `TaskScore` CRD: one immutable object per scoring event, deterministically
  named from (task, source kind, actor, raw signal) so redelivery is a no-op and
  retraction deletes exactly what the addition created.
- Result-artifact identity persisted as **labels** (indexable) on the Task and
  mirrored onto the TaskRecord, which outlives the Task's TTL.
- A shared normalization package; per-source adapters map to `Positive` /
  `Negative` / `Neutral` and preserve `raw`.
- Aggregation into `TaskRecord.status` plus Prometheus counters keyed by
  spawner, verdict, and source kind.

## Suggested order

1. Persist result-artifact identity as labels; add a TaskRecord status
   subresource. Nothing else works before this.
2. `TaskScore` CRD plus the controller that aggregates and emits metrics.
3. Slack reaction adapter — smallest diff, highest signal volume, lands on the
   agent's result message with no new UI.
4. GitHub `pull_request_review` adapter, reusing `GitHubCommentPolicy` for actor
   trust: the machinery that gates who may *trigger* an agent is exactly what
   should gate who may *score* one.
5. Shared `scoring:` config block across the remaining triggers.
6. Outcome signals (merge / revert / close) as distinct source kinds.

## Things that will bite us

- **Sparsity.** Most tasks will get zero scores. Any rate we display needs a
  minimum-n guard or it is noise presented as data.
- **Selection bias.** People react to surprisingly-good and egregiously-bad,
  never to *fine*. Reaction rate is not quality rate, and the aggregate will
  read better or worse than reality depending on channel culture.
- **The metric becomes the target.** Once per-spawner scores are visible there
  is pressure to tune agents toward reaction-bait phrasing. This is the
  strongest argument for weighting the outcome axis over the reception axis in
  anything we actually act on.
- **Trust and abuse.** Anyone who can react in a channel the bot is in can 👎.
  Gate or weight this via the existing comment-policy machinery rather than
  inventing a second permission model.

knechtionscoding:
#### What type of PR is this?

/kind feature

Adds a `TaskScore` API that records how actors outside the cluster judged an agent's result, and wires Slack emoji reactions up as the first source. A Task's terminal phase only tells us the process exited cleanly — it says nothing about whether the answer was useful. The people who know are outside the cluster and are already expressing that judgement; we were throwing it away.

This is the foundation, not the whole design. It deliberately ships the parts every version of result scoring needs, so later sources are adapters rather than rework.

**Correlation (the actual blocker).** Reporting was outbound-only. The Slack result message timestamp lived in an in-memory map keyed by Task UID and died with the pod, and the GitHub comment ID was an annotation, which `List` cannot filter on — resolving an inbound reaction would have meant reading every Task. The terminal Slack report now persists the result message's channel and timestamp as **labels**, so resolution is a single label-selector `List`. Only terminal phases are labelled: the "accepted" ack posted while a Task is still running is not a result, so reacting to it does nothing.

Tasks are TTL-deleted while scores keep arriving, so the labels are mirrored onto the TaskRecord. The two writers live in different processes and genuinely race — the reporter in `kelos-slack-server`, the record writer in `kelos-controller` — and there is an interleaving where a single pass loses the labels entirely: the reporter patches a record that does not exist yet (one attempt only, since it does not revisit a Task after recording the reported phase) while the controller creates the record from a cached Task revision predating the label write. The controller therefore backfills the labels when it finds the record already present, which the Task label write itself triggers a reconcile for. Regression-tested in both directions. Both writers use merge patches, so neither can conflict with the other on `resourceVersion` — which means both roles need `patch` on `taskrecords`, not `update`; a merge patch is an HTTP PATCH and `update` does not authorize it. The backfill is best-effort: its failure is logged rather than returned, so a scoring-correlation label can never fail budget accounting or hold up Task TTL cleanup, and `createTaskRecord` stays idempotent as its callers assume.

**`TaskScore`** is one immutable object per scoring event, not a counter on the TaskRecord: several actors can score the same result, scores arrive after the Task is gone, and append-only writes avoid the read-modify-write conflict storms that motivated making TaskRecord immutable. Each object is named from a hash of (Task UID, source type, actor, signal) — deliberately not the verdict — which makes the write path idempotent under Slack's at-least-once delivery and lets a retraction delete exactly the object its addition created.

**`internal/scoring`** is where normalization lives, and the only place it lives. Every source maps its own vocabulary onto a verdict there and records its own signal name alongside, so verdicts can be re-derived if a mapping turns out wrong. Per-trigger filtering has already diverged across the GitHub, Linear, generic, and Slack sources; this is structured to avoid repeating that.

**Slack reactions** are opt-in per spawner via `when.slack.scoring.reactions`, a `listType=map` list keyed by reaction name: 
```yaml
scoring: reactions:   - name: "+1"     verdict: Positive   - name: "-1"     verdict: Negative
```

A keyed list rather than parallel `positive`/`negative` lists makes the same emoji in two verdicts unexpressible (the API server enforces key uniqueness), so there is no precedence rule to define or overlap validation to write. It also keeps per-reaction options open — a bare enum map value could never later become an object without a scalar↔object kind change on a shipped field. Reactions not listed are ignored, so unrelated emoji in a busy channel do not dilute the aggregate. A reaction is evaluated against the mapping of the spawner that *created* the Task,
so one spawner's config cannot score another's results.
 
Retraction resolves the score by name and consults neither the verdict mapping nor any spawner, and is deliberately *not* gated on scoring being enabled somewhere. A reaction can be withdrawn after its emoji was dropped from the config, after the owning spawner was deleted, or after scoring was disabled cluster-wide. That last case is why the gate had to go: disabling scoring is a routine config edit, and a retraction event is transient — dropping one leaves the score counted forever with nothing to converge it. Recording still short-circuits when nothing scores, since it needs a spawner to supply the mapping.

**Aggregation** reconciles TaskRecords rather than TaskScores, watching scores and mapping them to the record named after their Task UID. A retraction deletes the score, and reconciling the deleted object could not recompute the summary — its `taskRef` is gone by then. The same reconcile adopts scores onto the record so they are reclaimed when its TTL expires.

**Metrics**: `kelos_task_score_recorded_total` and `kelos_task_score_retracted_total`, incremented only on real creates and deletes so redelivery does not inflate them. Actor is deliberately not a label — it is unbounded, and per-person scoreboards are not what this measures.

**API naming, bounds, and identity.** Two names follow Kubernetes convention rather than convenience. The source discriminator is `source.type`, not `source.kind` — `kind` denotes an object kind (`OwnerReference.Kind`) while union discriminators are `type` (`Condition.Type`), and `TaskRecord.spec.type` already sets the spelling here. The opaque locator is `source.uri`, not `source.ref` — `ref` means an object reference, and it would sit two fields from `spec.taskRef`, which is a structured one.

Every free-form string is bounded (`actor`/`signal` ≤ 256, `uri` ≤ 1024), since adding `MaxLength` later is a tightening change that would reject existing objects. Reaction names are validated against the alphabet Slack itself uses (`^[a-z0-9_+-]+$`): `White_Check_Mark`, `thumbs up`, a literal 👍, and `+1::skin-tone-4` are all things Slack never sends, so accepting them would record no scores with nothing logged. Note the pattern permits a leading `+`/`-`, without which the `+1` and `-1` aliases would be rejected; there is an envtest for that.

The fields that form a score's identity are constrained so the "one score per actor per distinct signal" contract is actually enforceable, since uniqueness rests only on `AlreadyExists`:

- `source.signal` is **required** and non-empty. Two scores differing only by a signal neither records would collide and the second be silently discarded.
- `source.actor` stays optional but rejects the empty string. Absence is meaningful: it means the source has no distinct actor — a machine-generated outcome signal — and identity then collapses to (Task, type, signal), which is the idempotency such a source wants.
- `taskRef.uid` is rejected when empty, not just when absent. An empty UID passes the required-key check but leaves a score that can never be summarized, adopted, or garbage-collected. The rule is CEL on `TaskScoreSpec` rather than `MinLength` on the shared `TaskReference`, because tightening that type would also tighten  `TaskRecord`, where the field already ships.

The verdict enum is `Positive`/`Negative` only, and `spec.spawnerName` was dropped because nothing read it — the `kelos.dev/taskspawner` label already carries provenance for selection. Both are additive to re-add.

**Five bugs found and fixed during review, worth a look:**

- **RBAC verb, twice.** The controller role had no write verb on `taskrecords` at all, so the label backfill was `Forbidden`; adding `update` then still did not authorize it, because both writers issue merge patches (HTTP PATCH). Both roles now grant `patch`, and the unused `update` is gone. The failure mode was compound: the labels never reach the record, so the whole "scores arriving after the Task's TTL still resolve" behaviour is dead, *and* the 403 propagated out of `createTaskRecord` so the terminal Task's reconcile failed forever, never reaching `applyTaskTTL`. Every unit test passed throughout — `fake.NewClientBuilder` does not enforce RBAC — which is why there is now a render test pinning these verbs (see Verification).
- **Thread spam.** The channel comes from an unconstrained annotation while label values are constrained, so an invalid value failed the whole update, leaving the reported phase unrecorded — the next cycle then re-posted the terminal message and failed again, forever. The labels are now gated on `validation.IsValidLabelValue`: scoring degrades, reporting stays correct.
- **Lifecycle coupling.** The backfill error propagated out of `createTaskRecord`'s `AlreadyExists` path, meaning a best-effort scoring label could block budget accounting and Task TTL. Now logged and swallowed.
- **Controller panic on the 1→0 score transition.** Retracting the *only* score on a Task made `summarizeScores` return nil while the status was still set, so the code fell through the no-op short-circuit and dereferenced the nil summary in the log line. Reachable by one person reacting and un-reacting. Found by adding the test for that specific branch — the 2→1 case that existed never nils the summary.
- **Conversion bypassed the v1alpha2 schema.** `kelos.dev/v1alpha2-slack-scoring` is user-writable, and CRD validation runs against the *request* version — v1alpha1, which has no scoring schema — so a hand-written annotation on a v1alpha1 create persisted a scoring block that violates the v1alpha2 constraints. An out-of-enum verdict then reached `TaskScore.spec.verdict`, where the enum *is* enforced, so every matching reaction failed at create with a logged error and no score. Restored entries are now validated against the same constraints the CRD applies and invalid ones dropped, preserving the "never block conversion" property. The validator lives in `internal/scoring` next to the matching logic so there is one Go-side source of truth; its test pins it against the CRD pattern's cases.

Deferred to follow-ups, per the minimum-viable-API convention: the GitHub adapters (`pull_request_review`, merge/close outcome signals), the shared `scoring:` block on the remaining triggers, and `actorTrust` weighting. `actorTrust` only earns its place once GitHub lands, where `GitHubCommentPolicy` supplies it; adding the field now would be speculative.

Also deferred, and worth tracking as issues rather than silently accepting:

- **Orphaned scores are unbounded — a deliberate deferral, not an oversight.** A  TTL field can be added additively later, so nothing here is schema-permanent. A  Task that finishes without reporting usage never produces a TaskRecord, so its scores are never adopted, never summarized, and never garbage-collected. The trigger is ordinary, not exotic: a Task that fails before its pod reports usage still posts a "failed" Slack message, and a 👎 on a failure is the archetypal negative score. `docs/reference.md` now calls this an unbounded growth path rather than stating retention as a fact, and gives operators the selector (`kelos.dev/task-uid`, plus empty `ownerReferences` as the marker for scores no record will reclaim). I considered owner-referencing the Task at creation and re-parenting to the record later, and rejected it: if the Task's TTL fires before the re-parent reconcile, a legitimate score is silently deleted, and retaining junk beats losing data. The root-cause alternative — writing TaskRecords unconditionally — changes existing TaskRecord semantics and needs its own decision.
- **The reserved labels are documented but not enforced.** `kelos.dev/slack-result-*` are now named in the reserved-labels table *and* in the `taskTemplate.metadata.labels` godoc, so `kubectl explain` shows the warning. A user can still set them, and a value shared across Tasks makes the message ambiguous, disabling scoring for it (logged, not silent). I deliberately did not add CEL enforcement: `taskTemplate.metadata.labels` already ships, so restricting keys is a tightening change, and `kelos.dev/taskspawner` is already prose-reserved on the same field —  CEL for only the new keys would be both non-compliant and inconsistent.

fixes #1586

**Four operational caveats worth your attention:**

1. **Reaction scoring needs Slack app changes that are not in this repo.** The app must grant `reactions:read` and subscribe to the `reaction_added` and  `reaction_removed` bot events. Without them Slack never delivers the events and no scores are recorded, **with nothing logged** — the CRD configuration alone looks correct and silently does nothing. Documented in the TaskScore reference section, but there is no in-repo Slack app manifest to change, so this is a manual step at deploy time.

2. **`kelos-slack-server` now caches TaskRecords, restricted by label.** Resolution falls back to TaskRecords, which starts a TaskRecord informer in that process. Left unrestricted, the default 30-day record TTL means a full copy of   every completed Task's record in the Slack pod, with every reaction event scanning all of it — and ungating retraction (above) makes that path hotter, since retractions now always resolve. The manager cache is therefore restricted to records carrying `kelos.dev/slack-result-ts`, which is exactly the set that is ever resolvable, so both footprint and scan stay proportional to results that can actually be scored. Safe because the only TaskRecord *read* in that process filters on the same label, and the label write is a patch, which bypasses the cache. Tasks are unrestricted — that informer already exists for the reporting loop.

3. **Reaction scoring is Slack-trigger-only.** `scoring` lives under `spec.when.slack`, and the handler only considers spawners with a Slack trigger, so a spawner that triggers from GitHub but reports to Slack cannot configure it. That is not reachable today — `AnnotationSlackReporting` is set in exactly one place, the Slack trigger path — but the schema does bake in the coupling. I kept the placement because `when.<source>.reporting` is the existing precedent for post-completion feedback config, and moving it now would be speculative; it is documented as a limitation instead. Say the word if you would rather have a sibling `spec.scoring` section before this ships.

4. **A Task that completes without reporting usage never gets a TaskRecord**, so it is only scoreable while the Task itself is alive, and any score it did collect is orphaned (see the deferred list above). This is inherited from  `createTaskRecord`'s existing `Status.Usage == nil` early return, not introduced here — but it means failed tasks, often the ones most worth a 👎, are the least likely to be scoreable later.
**Points I would most like a second opinion on:**

- The budget label snapshot interaction in `taskRecordLabels`. The snapshot *replaces* the Task's live labels and is taken at admission, long before a result exists, so the correlation labels are merged on top rather than inherited. The snapshot stays authoritative for budget matching, so labels added after admission cannot retroactively change which budgets a record counts against. Both behaviours are covered by tests, but this is the subtlest part of the change.
- Reconciling TaskRecord from two controllers (`TaskRecordReconciler` for TTL GC, `TaskScoreReconciler` for summaries, distinguished by `Named("taskscore")`). The alternative was folding summaries into the TTL reconciler, which mixes two unrelated concerns. Happy to go the other way if you would rather have one.
- `ctrl.SetControllerReference` sets `blockOwnerDeletion`, so I added a `taskrecords/finalizers` RBAC marker to match the existing `tasks/finalizers` and `taskspawners/finalizers` markers. Only matters in clusters running `OwnerReferencesPermissionEnforcement`, but the project already accounts for it elsewhere. The controller role does *not* get `delete` on taskscores — retraction happens in `kelos-slack-server` and reclamation via garbage collection.
- `status.scores` keeps explicit `positive`/`negative`/`total` rather than a `byVerdict` map. Status is additive-safe, so a third verdict just adds a field, and explicit fields keep the print columns and JSONPath simple. `total` is derivable today, and is kept as a stable denominator that stays correct under version skew — CRD upgraded, controller not — where a newer verdict is admitted and only `total` counts it.
- The field name `status.scores` holds a `TaskScoreSummary`, so it could read as a list of the scores themselves. I kept `scores` because `status.scores.positive` and `.total` read naturally and the type name already says what it is, and a real list will never go there (unbounded). Easy to settle either way before merge, since the name is permanent.
- Aggregation selects scores by the `kelos.dev/task-uid` label, which nothing can validate against `spec.taskRef.uid` (CEL cannot reach `metadata.labels`). `summarizeScores` now treats the spec as authoritative and skips any score whose `taskRef.uid` does not match, so the label is a prefilter rather than a correctness dependency. A field index on `spec.taskRef.uid` would remove the label from the path entirely — happy to do that instead if you prefer.
- The reserved label key constants live in `internal/scoring` while `AnnotationWorkerAssignedTask` and friends live in `api/v1alpha2`. Moving only the new ones would split the set, since `taskbuilder.SpawnerLabel` sets the opposite precedent — I'd rather move all of them in a standalone cleanup than bundle a half-move here.

**Placement decision, recorded permanently.** `scoring` sits under `spec.when`, which is documented as trigger conditions, while scoring is an observation of an outcome. The `Scoring` godoc now states the decision explicitly rather than leaving it implicit: config is per-trigger, matching the existing `when.<source>.reporting` fields, and the consequence is accepted permanently — only Slack-triggered spawners can score Slack reactions, because only they deliver a Slack result to react to. A source-independent scoring section would have to be a new field, not a move of this one. Flagging it as the decision I'd most want overridden before merge if you disagree, since it cannot be revisited.

**API compatibility:** all changes are additive. TaskRecord gains a status subresource and two priority-1 print columns; nothing writes its status today, so no client's writes are newly dropped — it is the one non-additive schema change in the diff and it is safe for that reason. `when.slack.scoring` is a new optional v1alpha2-only field, preserved through a v1alpha1 round-trip in an annotation following the existing `nameTemplate` and context-GitHub-App-auth pattern — without which a v1alpha1 read-modify-write would silently switch scoring off. No existing field gained tightened validation.

RBAC verbs are the one thing no test using the fake client can catch, so the new rules are pinned by a render assertion (`TestRender_ScoringRBACVerbs`) covering both roles — including a negative assertion that the controller does *not* get `delete` on taskscores. I verified it by reverting `patch` to `update` and watching it fail.

`taskscores.kelos.dev` had to be added to three hand-maintained lists: `kelosCRDNames` in `internal/cli/install.go` (without it `kelos uninstall` orphans the CRD), the chart CRD template generator in `hack/update-install-manifest.sh`, and the "Adopt Existing CRDs Into Helm" loop in the chart README (without it, that path leaves the CRD without Helm ownership metadata and the next `helm upgrade` fails on it). Only the first had a test, so I added `TestChartREADMEAdoptionLoopListsEveryCRD`, which derives the expected names from the chart templates and fails if the README omits any — verified by removing the line and watching it fail.

**Verification:** `make verify`, `make test` (32 packages), and `make test-integration` (167 envtest specs + 12 install specs) all pass.

Unit tests cover verdict mapping including skin-tone variants on both the event and the config side, resolution via Task and via TaskRecord fallback plus the ambiguous-match error, record idempotency under redelivery, retraction leaving other actors' scores intact, retraction after the emoji was dropped from the mapping and after the owning spawner was deleted and after scoring was disabled entirely, the reaction happy path end to end, summary recomputation after retraction, adoption, that an unchanged reconcile does not rewrite the record status, label persistence from both writers, the label-backfill interleaving above, the invalid-channel guard (labels skipped, phase still recorded, no re-post), the chart README CRD list, the scoring RBAC verbs in both rendered roles, the 1→0 summary transition, a mislabelled score being excluded from another Task's counts, invalid entries being dropped from a restored scoring annotation, and the conversion round-trip. The two counters are pinned too: redelivery does not inflate `recorded_total`, a retraction is counted with the verdict read from the stored object, and the documented empty-verdict fallback holds when the score is
unreadable.

New envtest specs cover what the fake client cannot evaluate — CEL and OpenAPI validation. On TaskScore: spec immutability, the verdict and source-type enums, the required and non-empty `taskRef.uid`, the required `source.type` and `source.signal` (absent and empty both rejected), empty `source.actor` rejected while absent `source.actor` is accepted, the length bounds, and crucially that a metadata-only update still succeeds against the spec-immutability rule — which is what both the adoption path and the reporter's record labelling depend on, and only a real API server can prove. On the spawner side: a well-formed mapping and the `+1`/`-1` aliases are accepted (so the rejection cases cannot pass vacuously on an unrelated schema error), and colon-wrapped names, skin-tone modifiers, uppercase names, spaces, literal emoji characters, empty and over-long names, unknown verdicts, a missing verdict, duplicate names, and more than 32 entries are all rejected.

The envtests earned their keep immediately: my first attempt at the `taskRef.uid` CEL rule used an empty-string literal that reached the CRD as a typographic quote, making the rule uncompilable and the CRD uninstallable. Unit tests were green; envtest failed at `BeforeSuite`. I also confirmed these specs genuinely execute by breaking one and watching it fail, rather than trusting the pass count.

```release-note
Added a TaskScore API recording how external actors judged an agent's result, with Slack emoji reactions as the first source. Map reactions on an agent's result message to verdicts under `TaskSpawner.spec.when.slack.scoring.reactions`, a list of `{name, verdict}` entries where verdict is `Positive` or `Negative` — for example `- name: "+1"` with `verdict: Positive`. Reaction names use Slack's own alphabet (lowercase letters, digits, `_`, `-`, `+`) without surrounding colons or a skin-tone modifier. Reactions that are not listed are ignored, and removing a reaction retracts its score. Scores are aggregated into `TaskRecord.status.scores` and exported as the `kelos_task_score_recorded_total` and `kelos_task_score_retracted_total` metrics. Reaction scoring is available on Slack-triggered spawners only, and requires the Slack app to grant the `reactions:read` scope and subscribe to the `reaction_added` and `reaction_removed` bot events.
```

---
## Summary by cubic
Adds a new `TaskScore` API to record how people judge agent results, starting with Slack emoji reactions. Scores are summarized on `TaskRecord.status.scores`, using persisted Slack message labels to resolve reactions even after Task TTL.

- **New Features**
  - New CRD `taskscores.kelos.dev`: one immutable object per scoring event; idempotent name from (task UID, source type, actor, signal); retraction deletes the exact score.
  - Slack reactions as first source: opt-in per spawner at `spec.when.slack.scoring.reactions`; only listed reactions count; names use Slack’s alphabet (e.g. `+1`, `white_check_mark`); removing a reaction retracts the score.
  - Correlation: result message `channel`/`ts` are written as labels on `Task` and mirrored to `TaskRecord`; invalid values are skipped to prevent repost loops; `kelos-slack-server` restricts its cache to records with the result label and retries briefly to resolve reactions posted near label writes.
  - Aggregation: a new controller updates `TaskRecord.status.scores` (positive/negative/total), adopts scores for GC, and adds `Positive`/`Negative` print columns; TaskRecord gets a status subresource.
  - Metrics/RBAC: exports `kelos_task_score_recorded_total` and `kelos_task_score_retracted_total`; both controllers and `kelos-slack-server` use `patch` on `taskrecords` (no `delete` on `taskscores`).

- **Migration**
  - Install the new CRD `taskscores.kelos.dev` (Helm templates and uninstall paths updated).
  - Update your Slack app: add `reactions:read` and subscribe to `reaction_added` and `reaction_removed`.
  - Configure spawners with `spec.when.slack.scoring.reactions` (max 32 entries). Example: `- name: "+1", verdict: Positive`.
  - Do not set `kelos.dev/slack-result-channel` or `kelos.dev/slack-result-ts` on `taskTemplate.metadata.labels`; these are reserved for correlation.
  - If you manage RBAC manually, grant `patch` on `taskrecords` (and `taskrecords/status` if applicable) to both the controller and `kelos-slack-server`.

<sup>Written for commit f54f434c337bfc733a74e8253345dd862cc5ed47. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1588?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


knechtionscoding:
@gjkim42 curious if you have thoughts here/if this is a reasonable thing?

gjkim42:
Not sure about this...
It doesn't have meaningful behavior eventually but storing the score.

I think it's not related to kubernetes. We can still do that without a kubernetes controller and store the score in task's metadata (annotations or labels), right?

knechtionscoding:
> Not sure about this... It doesn't have meaningful behavior eventually but storing the score.
> 
> I think it's not related to kubernetes. We can still do that without a kubernetes controller and store the score in task's metadata (annotations or labels), right?

We could. I debated that. The problem? Is that it gets removed when the TTL for the task expires. So the score has to be given and collected prior to the TTL. That might be fine? Wanted to put up the idea.

knechtionscoding:
@gjkim42 I can make this an annotation/label based method. I think exposing it as a metric would also satisfy the requirements. Basically that when a user reacts positively or negatively to it we increment a metric for positive, negative, or no outcome measured, and then attach it to either a task or task spawner. That allows for measuring what specifically I want without adding too much? does that sound reasonable?

gjkim42:
Having score metrics through annotations seems acceptable.
I just want to avoid scraping external score by kelos.

knechtionscoding:
> Having score metrics through annotations seems acceptable.
> I just want to avoid scraping external score by kelos.

@gjkim42 is it acceptable if Kelos responds to the events? We don't have to scrape, but building a separate system that listens or scrapes github PRs, slack reactions, etc seems silly when Kelos can listen for reactions to comments/PR status/slack reactions. All of those can happen and write to the annotation of a task without extending a CRD but also without forcing a separate external system to read the data or rebuild the listener and mappings.

knechtionscoding:
@gjkim42 bump here

knechtionscoding:
@gjkim42 bump here if you have a chance

gjkim42:
I think still it's better to separate that from the kelos.

PR #1620, opened 2026-08-08
Report runtime status from Claude Code and OpenCode Sessions

gjkim42:
#### What type of PR is this?

/kind feature

Only the Codex provider emitted `runtime.status` events, so the session status bar (terminal and web) and `status.model` showed the model, cumulative token usage, and context use for Codex Sessions only; claude-code and opencode Sessions displayed just the static basics.

This brings both providers to parity with Codex:

- **Claude Code**: records the model from the `system`/`init` event and from assistant messages, accumulates input/output tokens from assistant message usage for live mid-turn updates (deduplicated by message ID, since usage repeats on every content-block event of a message), reconciles totals from each result event's session-cumulative `usage` (authoritative — it also covers subagent and internal API calls that never appear as assistant events in the stream), derives context use from the latest message's usage, and reads the context window from the result event's `modelUsage`.
- **OpenCode**: records the model (`provider/model`) from assistant `message.updated` events, sums token usage from `step-finish` parts (deduplicated by part ID) including cache and reasoning tokens (OpenCode reports `tokens.input` without cache tokens and `tokens.output` without reasoning tokens, so both are added back), derives context use from the latest step, and resolves context windows once at startup from the server's `/config/providers` model limits (best-effort; a failed lookup only omits the context percentage).

Both providers now implement `runtimeStatusSnapshot()`, so the server seeds their status at construction and they emit it at each turn start, matching the Codex flow. As a result `status.model` is also populated with the actual runtime model for claude-code and opencode Sessions after the first turn.

Docs are updated: context use and token counts are no longer documented as Codex-only; weekly limit remaining stays Codex-only because neither Claude Code stream-json nor the OpenCode server API exposes rate-limit data.

N/A

- The weekly limit segment intentionally remains Codex-only; the status bar simply hides it when absent.
- Token-accounting semantics were verified against the pinned agent versions: Claude Code 2.1.223 (result events carry process-cumulative `usage`/`modelUsage` aggregated across all API calls) and OpenCode 1.18.11 (`getUsage` computes `output` as `outputTokens - reasoningTokens` and `input` without cache tokens).
- `TestOpenCodeProviderStreamsOrderedEvents` gained a leading `runtime.status` event in its expected sequence because OpenCode now emits status at turn start, as Codex already did.
- New tests: `TestClaudeProviderRuntimeStatusMapping` (including result-usage reconciliation exceeding the visible assistant-message sums) and `TestOpenCodeProviderRuntimeStatusMapping` (including a nonzero-reasoning step), mirroring the existing `TestCodexRuntimeStatusMapping`; the fake OpenCode server now serves `/config/providers`.

```release-note
claude-code and opencode Sessions now report the runtime model, cumulative input/output tokens, and context use in the session status bar and `status.model`, matching codex Sessions. Weekly limit remaining is still reported by codex Sessions only.
```

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -244,10 +244,11 @@ waiting-for-input, and interrupting progress with elapsed time. After a
 completed turn, both interactive and plain terminal output show a `Worked for
 ...` separator when the duration is known. A separate status bar beneath the
 composer shows the Session name, agent type, model and effort when available,
-working directory, git branch, and associated pull request number. Codex
-Sessions also show reported context use, weekly limit remaining, and cumulative
-input and output tokens. Less important status-bar items are omitted as the
-terminal narrows. The model is the same runtime-reported value persisted in
+working directory, git branch, and associated pull request number. Sessions
+also show reported context use and cumulative input and output tokens once the
+agent reports usage; Codex Sessions additionally show weekly limit remaining.
+Less important status-bar items are omitted as the terminal narrows. The
+model is the same runtime-reported value persisted in
 `status.model`. Web chat is served by the optional shared
 `kelos-session-server`; it shows the same live runtime details beneath its
 composer. It shows connection status separately from working, waiting-for-input,
--- a/internal/sessionruntime/claude.go
+++ b/internal/sessionruntime/claude.go
@@ -57,6 +57,14 @@ type ClaudeProvider struct {
 	blockText         map[int]*strings.Builder
 	interrupted       bool
 
+	statusMu sync.Mutex
+	model    string
+	usage    *RuntimeUsage
+	// countedMessages holds assistant message IDs whose usage was already added
+	// to the totals; Claude Code repeats a message's usage on every content-block
+	// event for that message.
+	countedMessages map[string]struct{}
+
 	sessionMu sync.Mutex
 	sessionID string
 	done      chan struct{}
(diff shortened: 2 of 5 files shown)


gjkim42:
Addressed both P2 findings from the Kelos Reviewer Agent in 81a9304 (squashed). I verified each against the pinned agent versions before changing code:

**Claude result usage (claude.go)** — Confirmed against the installed `@anthropic-ai/claude-code@2.1.223` binary: the persistent stream-json engine builds every result event from process-global cumulative state (`total_cost_usd`, `usage`, `modelUsage` are all derived from the same per-model accumulator, with `usage` computed by summing across all models), so it covers API calls that never surface as assistant envelopes in the root stream. The provider now replaces its running totals with each result event's usage; per-assistant-message accumulation is kept only for live mid-turn updates and context-use tracking. Zeroed synthetic result usage is ignored so it cannot wipe the totals. Test covers a result whose usage exceeds the visible assistant-message sums.

**OpenCode reasoning tokens (opencode.go)** — Confirmed against `sst/opencode` v1.18.11 source (`session.ts` `getUsage`): `tokens.output = outputTokens - reasoningTokens`, with reasoning reported separately (and `tokens.input` excludes cache tokens, so the existing cache add-back was already correct). The decoder now reads `tokens.reasoning` and includes it in output, total, and context tokens. Test covers a nonzero-reasoning step.

`make verify` and `make test` pass; commits squashed to one.

PR #1646, opened 2026-08-12
Add excludeFilters to generic webhooks

### LINKED ISSUE #1635: Add `excludeFilters` to generic Webhooks
Hi!

Since there are only positive filters on generic Webhooks it's hard to implement something like `excludeAuthors` from `githubWebhook`. It would be nice to either have a `exclude: true` option in the filters or a separate `excludeFilters` Property in `spec.when.webhook` (Task- and Session Spawners)

knechtionscoding:
#### What type of PR is this?

/kind feature

Generic webhooks (`spec.when.webhook`) only supported positive match
conditions, so there was no way to express the exclusions that
`githubWebhook` gets from `excludeAuthors` / `excludeLabels` — a delivery
either matched every entry in `filters` or was skipped.

This adds `spec.when.webhook.excludeFilters`, reusing the existing
`GenericWebhookFilter` type (so `field` plus exactly one of `value` or
`pattern`, enforced by the same CEL rule as `filters`).

Semantics: a delivery triggers a Task only when it matches every entry in
`filters` **and no** entry in `excludeFilters` (OR semantics across exclude
filters, matching `excludeLabels` / `excludeBodyPatterns` elsewhere in the
API). An exclude filter whose `field` is absent from the payload does not
match, so it never excludes the delivery — the inverse of a positive filter,
where a missing field fails the match.

Also included: reference docs rows, an `excludeFilters` section and example
config in `examples/13-taskspawner-generic-webhook/`, the generic-webhook
section of `docs/integration.md`, and regenerated CRDs and deepcopy.

Fixes #1635

Two design/scope decisions worth a look:

- **A separate `excludeFilters` list rather than a per-filter `exclude: true`
  flag.** The issue offered both. A separate list matches the title and the
  established convention in this API (`excludeLabels`, `excludeAuthors`,
  `excludeBodyPatterns`, `slack.excludePatterns`), and it keeps the AND
  semantics of `filters` from getting tangled with the OR semantics of
  exclusion within one list.
- **v1alpha2 only.** Per the repo convention on maintaining API changes in
  the latest version. Like the existing `githubWebhook` `excludeBodyPatterns`
  field, it is dropped by a v1alpha1 round-trip; no preservation annotation
  was added, consistent with how that field was handled.

SessionSpawner is intentionally untouched: its `when` supports only
`githubWebhook`, so there is no generic webhook there to filter. The issue
mentions Session Spawners and the triage comment flagged this as needing a
maintainer decision — adding generic webhook support to SessionSpawner is a
separate feature, not part of this change.

One behavior change beyond the new field: `jsonpath.Get` returns an error
both for an absent payload field and for a malformed expression, and the
filter code collapsed both into "does not match". For an `excludeFilters`
entry that fails open — an API-accepted typo (empty `field`, malformed
JSONPath) would drop the exclusion and spawn a Task the operator meant to
reject. Filter evaluation now separates the two the way `ExtractFields`
already does: an absent field stays a plain non-match, while a syntax error
is returned as a configuration error, which skips that spawner for the
delivery and logs it. This applies to `filters` too, where the old behavior
was merely silent rather than unsafe.

Tests: `TestMatchesGenericExcludeFilters` covers value/pattern matches, OR
semantics, missing fields, invalid regex, and malformed JSONPath.
`TestMatchesGenericFilters_MalformedJSONPath` and
`TestMatchesGenericFilters_MissingFieldIsNotAnError` pin the
error-vs-non-match distinction for both functions.
`TestGenericServeHTTP_ExcludeFilters` is a handler-level table test asserting
Task creation is actually suppressed on a match, still happens when the
excludes do not match, and is suppressed (fails closed) when an exclude
expression is malformed.

```release-note
TaskSpawner generic webhooks (`spec.when.webhook`) now support `excludeFilters`, which reject a delivery when any entry matches. A delivery spawns a Task only when it matches every entry in `filters` and no entry in `excludeFilters`.
```

---
## Summary by cubic
Adds `excludeFilters` to generic webhooks to reject deliveries; a Task now spawns only when all `filters` match and no `excludeFilters` do. Also treats malformed JSONPath in either list as a configuration error that skips the spawner for that delivery (previously collapsed into a non-match).

- `spec.when.webhook.excludeFilters`: same `GenericWebhookFilter` shape as `filters` (`value` or `pattern`), OR semantics across excludes, evaluated after `filters`; missing fields don’t match; invalid regex returns an error.
- v1alpha2-only; dropped on v1alpha1 round-trip.
- Updated CRDs/docs/examples and regenerated deepcopy; handler evaluates excludes; tests cover value/pattern matches, missing fields, invalid regex, malformed JSONPath, and handler skip-on-error.

<sup>Written for commit 39c03cff82e48d094c39587c772a25868d679b6a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1646?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gjkim42:
Let's ignore this.
I don't want to support roundtrip for alpha apis.

/kelos review

PR #1655, opened 2026-08-18
Support idlePolicy in the Session create API

knechtionscoding:
#### What type of PR is this?

/kind feature

`Session.spec.idlePolicy` (`suspendAfterSeconds` / `deleteAfterSeconds`) is fully supported by the
CRD and enforced by `SessionReconciler`, but `POST /api/sessions` could not carry it:
`createSessionRequest` lacked the field and `createSession` built `SessionSpec` without it. Because
`decodeJSON` uses `DisallowUnknownFields`, sending an `idlePolicy` key returned 400 rather than
being silently ignored — so clients had no way to configure idle suspension or deletion at Session
creation time.

This adds `IdlePolicy *kelos.SessionIdlePolicy` to `createSessionRequest` and passes it through to
the created `SessionSpec`.

`POST /api/sessions/apply` is not a substitute for callers that need this: it is an upsert and so
never returns 409, and clients rely on that 409 to detect and adopt orphaned Sessions.

Behavior is otherwise unchanged:

- No default is applied — omitting `idlePolicy` still produces a Session with no idle policy.
- No handler-side validation was added. `Minimum=0` and the CEL rule
  (`suspendAfterSeconds < deleteAfterSeconds`) already live on `SessionIdlePolicy`.
- No API or codegen changes: `SessionSpec.IdlePolicy` and its deepcopy already exist.

N/A

Tests cover a full policy round-trip, the omitted-field case (`spec.idlePolicy` stays nil), and a
partial policy with only `deleteAfterSeconds` set.

No CEL-rejection test is included. The session server test harness builds its client with
`fake.NewClientBuilder()`, which does not evaluate CEL rules, so such an assertion would pass
vacuously regardless of the handler's behavior.

Out of scope: `createSession` maps both `IsAlreadyExists` and `IsInvalid` to 409, so an invalid
idle policy surfaces to the client as a conflict rather than a validation error. That conflation
predates this change and is untouched here.

```release-note
The session server create API (`POST /api/sessions`) now accepts an `idlePolicy` field, so clients can configure idle suspension and deletion when creating a Session.
```

---
## Summary by cubic
Accepts `spec.idlePolicy` in POST /api/sessions so clients can configure idle suspension/deletion at creation. Previously the create API rejected `idlePolicy` with 400; now it passes the field through. Omitting it still creates no policy.

- Request shape unchanged (spec fields remain inline); only `spec.idlePolicy` is newly accepted.
- No new handler validation; CRD Minimum/CEL rules still enforce constraints. Invalid specs continue to return 409; the apply endpoint remains an upsert.
- Tests cover idlePolicy round-trip and omit/partial cases.

<sup>Written for commit 015b80af272440c3b36953e88cf825f9bee88818. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1655?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -4,7 +4,9 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/x509"
 	"encoding/base64"
+	"encoding/pem"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


internal/sessionserver/server.go
gjkim42, line 144: Can we just include Session object here?

knechtionscoding: Done

PR #1660, opened 2026-08-19
fix(slack): stop retrying replies rejected with cannot_reply_to_message

knechtionscoding:
#### What type of PR is this?

/kind bug

When Slack rejects a thread reply with `cannot_reply_to_message` (for example, when the thread is closed or the bot cannot reply to it), the `kelos-slack-server` reporting cycle currently retries the post forever. Because the failed post prevented the task's `kelos.dev/slack-report-phase` annotation from being persisted, the reporter re-posted on every 30s tick — observed in production as two self-triggered tasks wedged in the `accepted` phase with 198+ failed attempts in a single hour.

This PR treats `cannot_reply_to_message` as a permanent API error:

- **Phase transitions** (`ReportTaskStatus`): log the failure once and still persist the report phase, so the reporting cycle stops retrying that reply on every tick. HTTP-level failures (timeouts, 5xx, rate limits) keep the existing retry behavior.
- **Progress snapshots** (`updateProgress`): mark the snapshot as posted on a permanent error so it is not re-attempted every cycle.

The permanent-error detection is isolated in a small `isPermanentSlackError` helper, living alongside the other `SlackReporter` helpers in `internal/reporting/slack.go`, so the classification is easy to extend to other permanent Slack API errors later.

N/A

- The new behavior is covered by unit tests:
  - `TestSlackTaskReporter_CannotReplyToMessageNotRetried` — permanent error: one attempt, phase annotation persisted, no further attempts on the next cycle.
  - `TestSlackTaskReporter_TransientReplyErrorStillRetried` — transient error: error surfaced, annotation not set, retry behavior unchanged.
  - `TestSlackTaskReporter_PermanentProgressErrorNotReattempted` — permanent error on a progress snapshot: single attempt, not re-attempted next cycle.
  - `TestIsPermanentSlackError` — classification of wrapped errors, other Slack API errors, HTTP status errors, and plain errors. Includes `rate_limited`, asserting rate limits stay transient and keep their existing retry behavior.
- `isPermanentSlackError` matches `slack.SlackErrorResponse` by value only. That is the shape slack-go v0.20.0 returns from `SlackResponse.Err()` (`misc.go:180`, `misc.go:195`), `Error()` has a value receiver, and no path in the library returns a pointer, so a `*SlackErrorResponse` branch would be unreachable.
- The permanent-error path uses `break` rather than an early return, so it falls through to the existing tail of `ReportTaskStatus`: `setActivityTarget` is skipped (`firstReplyTS` is empty), and terminal phases still reach `clearProgressCache`/`clearActivityState` before the phase is persisted.
- Verified: `go build ./...`, `go vet`, `gofmt`, and `go test ./internal/reporting/` all pass. The new tests fail against the pre-fix code. The repo-wide failures (`internal/controller` entrypoint tests, `test/e2e`, `test/integration`) are pre-existing environment issues (kubebuilder envtest binaries and a git sandbox limitation) that fail identically without this change.
- `make test-integration` requires envtest binaries not available in the sandbox environment.

```release-note
Slack task reports that Slack permanently rejects with cannot_reply_to_message are no longer retried on every reporting cycle, preventing repeated duplicate posts.
```

---
## Summary by cubic
Stop retrying Slack thread replies rejected with cannot_reply_to_message by classifying them as permanent API errors. This prevents 30s retry loops, wedged tasks, and duplicate posts by persisting the report phase and suppressing progress posts.

- Adds `isPermanentSlackError` that matches `slack.SlackErrorResponse.Err == "cannot_reply_to_message"`; HTTP failures and rate limits remain retryable.
- Phase transitions: on permanent error, log once with channel and thread timestamp, persist `kelos.dev/slack-report-phase`, and set a per-task progress block so the next cycle does not attempt progress posts.
- Progress updates: if blocked, return before reading pod logs; on permanent error, set the block and log channel and thread timestamp on failures.
- The progress block is in-memory: a restart/leader failover may trigger one extra rejected post before re-blocking; it clears via `clearProgressCache` and is swept in `SweepProgressCache`.
- Unit tests cover permanent vs transient paths, progress suppression, and error classification. No config or migration changes.

<sup>Written for commit b84f50d7a2f7632fe3d045a5b825dbcdad46166f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1660?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/reporting/slack.go
+++ b/internal/reporting/slack.go
@@ -3,6 +3,7 @@ package reporting
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"sync"
 	"unicode/utf8"
@@ -76,6 +77,27 @@ func (r *SlackReporter) UpdateMessage(ctx context.Context, channel, messageTS st
 		return fmt.Errorf("updating Slack message: %w", err)
 	}
 	return nil
+}
+
+// isPermanentSlackError reports whether err is a Slack API error that will
+// never succeed if retried, such as replying to a message Slack does not
+// allow replies to ("cannot_reply_to_message"). HTTP-level failures
+// (timeouts, 5xx) and rate limits are transient and remain retryable.
+//
+// slack-go returns SlackErrorResponse by value, with the API error code in
+// its Err field.
+//
+// The set is deliberately limited to cannot_reply_to_message. Other codes
+// that are arguably permanent (is_archived, channel_not_found, invalid_blocks,
+// msg_too_long) still retry, so suppressing them is a behavior change that
+// belongs in its own change with its own tests. Add them here when that
+// happens.
+func isPermanentSlackError(err error) bool {
+	var respErr slack.SlackErrorResponse
+	if !errors.As(err, &respErr) {
+		return false
+	}
+	return respErr.Err == "cannot_reply_to_message"
 }
 
 // contextBlock returns a context block displaying the task name.
--- a/internal/reporting/slack_test.go
+++ b/internal/reporting/slack_test.go
@@ -3,6 +3,8 @@ package reporting
 import (
 	"context"
 	"encoding/base64"
+	"errors"
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


gjkim42 (approved):
Lgtm, thanks!

PR #1692, opened 2026-08-25
Fail Claude Code runs that end with a degenerate final message

knechtionscoding:
#### What type of PR is this?

/kind bug

Claude Code can finish a run cleanly — `subtype: success`, `is_error: false`,
`stop_reason: end_turn`, `terminal_reason: completed` — after doing substantial
real work, and still emit markup fragments instead of an answer as its final
message. A Slack-facing agent did exactly this in production, ending a
41-turn run with:

```
`stale=False</li>
</ul>
</section
</section>
```

The completion gate in `kelos-capture` only inspects the termination fields, so
this passes, the Task is recorded as Succeeded, and the fragments are reported
verbatim.

This adds a content predicate. A result is classified `ResultIncomplete` when
all of the following hold:

1. it would otherwise be `ResultCompleted` (an already-failing run is not
   second-guessed),
2. `num_turns >= 10` — the agent did substantial work, and
3. the trimmed final message is under 200 characters.

`ResultIncomplete` is reused rather than adding a new status, because it already
makes `kelos-capture` exit non-zero. The Job's existing unconditional
`backoffLimit: 1` turns that into a rerun, and `isJobFailed` waits for the
`JobFailed` condition rather than `job.Status.Failed > 0`, so the Task stays
`Running` across the retry. No new retry machinery is needed.

`kelos-capture` also emits a `degenerate: true` output key on the failing
attempt (`run()` emits the marker block even when the exit code is 1), so
downstream consumers can distinguish this failure from any other without
duplicating the heuristic.

N/A

**The predicate measures characters, not tokens.** `usage.output_tokens` on the
result line appears to be cumulative over the session, not per-final-turn, so a
run with 40 tool calls carries a large `output_tokens` even when the final
message is junk — a token-based floor would silently never fire. `result` is
unambiguously the final message, so measuring its length sidesteps the question.
Length is counted in runes, not bytes, so the floor means the same thing for a
non-ASCII answer as for an ASCII one.

**The `degenerate` marker is gated on the classification, not the length
floor.** `IsDegenerateOutput()` is independent of `Status()`, so an explicit
error result (`subtype: error_max_turns`) whose short error text trips the same
floor would otherwise be labelled degenerate. That is a different failure, and
mislabelling it would hide the real reason from downstream consumers, so the
marker is only set when the result is `ResultIncomplete` *and* degenerate.

**This is not provider-side truncation.** FireRouter was probed directly against
`deepseek-v4-flash-0731`: `max_tokens: 8` returns `stop_reason: max_tokens` and
`max_tokens: 512` returns `end_turn`. `stop_reason` is forwarded faithfully, so
the existing `max_tokens` guard already catches genuine truncation. The garbled
run came back as a real clean `end_turn`; content is the only remaining signal.

**A false positive is cheap.** If only the first attempt is misjudged, it costs
one retry and is invisible to the user — the Task stays `Running` and Slack
shows only the progress message. If both attempts are misjudged, the Task goes
`Failed` and Slack posts the error header with `results["response"]` still
attached, so the user still gets their answer. A false positive never loses an
answer, which is the argument for erring toward detection.

**The thresholds (10 turns, 200 chars) are deliberately conservative and not
yet calibrated.** There is no metric for final-message length today; the
distribution has to come from `task.Status.Results["response"]` on completed
Tasks. `kelos_task_output_tokens_total` is labeled by spawner as of `c2faf5f1`,
so the token distribution per spawner is already queryable, but per the note
above tokens are the wrong measure here. Intent is to ship conservative and
tighten once the distribution is known.

Backward compatibility: older Claude Code versions that omit `num_turns` leave
it at zero, so the predicate never fires for them — matching how `Status()`
already tolerates empty reason fields. There is an explicit test for this.

Tests cover the real garbled sample after a high turn count, a legitimate brief
answer after two turns (`"Fixed in v0.472."` — the case the predicate must not
break, since the general spawner answers casually all the time), a long
coherent answer after many turns, the `charFloor` and `turnFloor` boundaries,
the missing-`num_turns` case, and that `degenerate: true` reaches the captured
outputs on a failing attempt.

`make test` and `make verify` pass.

```release-note
Claude Code runs that terminate cleanly but return a final message too short to be an answer after many turns are now treated as failures rather than successes, so the Task's existing retry reruns the agent instead of reporting garbled output. `kelos-capture` emits a `degenerate: true` result key on such an attempt.
```

### DIFF
--- a/docs/agent-image-interface.md
+++ b/docs/agent-image-interface.md
@@ -266,6 +266,13 @@ entrypoint preserves Claude Code's own non-zero exit code when present;
 otherwise, it propagates the capture failure so the Task is not reported as
 successful.
 
+A run that terminates cleanly but returns a final message too short to be an
+answer after many turns is also treated as a failure, since Claude Code reports
+such a run as a success. `kelos-capture` emits `degenerate: true` alongside the
+usual outputs in that case and returns non-zero, so the Job's retry reruns the
+agent. The `response` output is still captured on the failing attempt, so the
+agent's final message is recorded in the Task's results either way.
+
 Also use `set -uo pipefail` (without `-e`) so the capture step runs even if
 the agent exits non-zero.
 
--- a/internal/capture/capture.go
+++ b/internal/capture/capture.go
@@ -20,7 +20,8 @@ const (
 
 // Run streams the agent's JSON output from stdin to stdout, accumulating
 // per-agent token usage in memory, then emits deterministic outputs
-// (branch, commit, PRs, token usage) between markers on stdout. It is
+// (branch, commit, PRs, token usage, and a degenerate marker when Claude
+// Code returned an unusable final message) between markers on stdout. It is
 // intended to be the right-hand side of a pipe from the agent process so
 // that no on-disk copy of the stream is required. It returns non-zero when
 // the stream cannot be processed or Claude Code reports an incomplete result.
@@ -95,7 +96,7 @@ func captureOutputs(r runner, usage map[string]string) []string {
 		}
 	}
 
-	for _, key := range []string{"cost-usd", "input-tokens", "output-tokens", "response"} {
+	for _, key := range []string{"cost-usd", "input-tokens", "output-tokens", "response", "degenerate"} {
 		if v, ok := usage[key]; ok {
 			outputs = append(outputs, key+": "+v)
 		}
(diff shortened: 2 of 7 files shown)


gjkim42 (commented):
/kelos claude-review

PR #1708, opened 2026-08-28
Match GitHub webhook events by pull request author

knechtionscoding:
#### What type of PR is this?

/kind api

Adds `pullRequestAuthor` as an ordinary matching criterion on `githubWebhook` filters, plus a source-level `excludeFilters` list that reuses the same criteria and the same matcher — the shape `GenericWebhook` already uses for `filters`/`excludeFilters`.

`githubWebhook.excludeAuthors` matches the event sender and nothing else. That makes it loop prevention — it suppresses the burst a bot emits about its own activity — but it cannot express "don't act on this bot's pull requests". Any human interaction with a bot-authored PR re-admits the event: taking it out of draft, reopening it, pushing to it, or commenting on it are all sent by a human.

Concretely, a spawner that lists a bot under `excludeAuthors` still fires on that bot's own PR:

| actor | event | outcome |
|---|---|---|
| `some-bot` | PR opened | dropped — sender is excluded |
| a human | `ready_for_review` | **matched** — sender is a human |

No webhook filter could express the author-based match before this change, because `GitHubEventData` carried no pull request author at all. This is a gap in the matcher, not a misconfiguration.

```yaml
githubWebhook:
  events: ["pull_request", "issue_comment"]
  excludeFilters:
    - pullRequestAuthor: "dependabot[bot]"
```

Changes:

- **API** (`api/v1alpha2/taskspawner_types.go`): `PullRequestAuthor` on `GitHubWebhookFilter` (`MaxLength=64`) and `ExcludeFilters []GitHubWebhookFilter` on `GitHubWebhook` (`MaxItems=20`), plus regenerated deepcopy and CRDs. SessionSpawner inherits both through the shared `*GitHubWebhook`.
- **Parsing** (`internal/webhook/github_filter.go`): `PullRequestAuthor` on `GitHubEventData`, from the typed payload for `pull_request` / `pull_request_review` / `pull_request_review_comment`, from `issue.user` for `issue_comment` when `issue.IsPullRequest()`, and from raw JSON for PR-bearing events with no typed case arm.
- **Matching**: one matcher serves both paths. `matchesAnyGitHubExcludeFilter` is evaluated in `MatchesGitHubEvent` before the accepting list, and again in `githubWebhookNeedsChangedFiles` so an excluded event skips the changed-files fetch.
- **Validation**: two CEL rules on `GitHubWebhook` — every accepting filter must set `event`; an exclusion rule must set at least one criterion and cannot use `filePatterns`.
- **Conversion** (`internal/conversion/taskspawner.go`): `excludeFilters` is v1alpha2-only, preserved across a v1alpha1 round-trip via `kelos.dev/v1alpha2-github-webhook-exclude-filters`, with the decoded rules re-checked against the same bounds before they are applied.
- **CLI**, **docs** (`docs/reference.md`, `docs/integration.md`) and examples 10 and 17.

N/A

- **Reshaped from the previous revision per @gjkim42's review.** It previously added a scalar `excludePullRequestAuthors`; that field is gone. Exclusion is now expressed as rules over the shared criteria, so the positive and negative paths cannot drift.
- **Semantics.** Criteria within a rule are ANDed; any matching rule rejects the event regardless of which accepting filter would have taken it; rules sit outside the OR-based accepting list. A rule may omit `event` to span every subscribed event type, which accepting filters may not do. A criterion absent from the event does not match, so `push`, `issues`, and `check_run` are never rejected by a `pullRequestAuthor` rule. All of this is covered by tests.
- **`filters` is now bounded at `MaxItems=50`, and this is the one thing worth your explicit sign-off.** Making `event` optional (so exclusion rules can omit it) shrinks the minimum `filters[]` item, which inflates the API server's estimated array length and pushed the **pre-existing** checks-reporting CEL rule 2.1x over the cost budget — the CRD would not install. Bounding `filters` restores the estimate. The alternative was splitting the criteria into a shared struct inlined by two filter types, keeping `event` required on the accepting one; that avoids touching `filters` but rewrites all 64 `GitHubWebhookFilter{...}` literals in the test suite and nests every future literal. I went with the bound as the smaller change, and I did not touch the pre-existing rule itself. It is a cap on a shipped field, so an in-cluster spawner with more than 50 filters would stop applying — the largest list anywhere in tree is 6. Say the word if you'd rather have the struct split, or a different cap.
- **The raw-JSON extraction is keyed on the payload, not on an event allowlist.** `pull_request_target` has no typed case arm and falls through to `default`, which previously read only `sender` and `action` — yet `handler.go` accepts the event and the docs list it as supported. Exactly five go-github event types carry a top-level `pull_request` object; three have case arms, and the two landing in `default` are `pull_request_target` and `pull_request_review_thread` — both PR events where matching the author is the desired behavior, and both covered by tests. Keeping it payload-keyed means a PR-bearing event GitHub adds later works without a code change, and `pull_request.user` is the PR author throughout GitHub's webhook schema.
- **`check_run` is deliberately out of scope.** Its `pull_requests[]` entries are minimal and carry no `user`, so the author would need an API fetch. The lazy-fetch machinery exists but is not extended here; the godoc, reference table, and CI-remediation example all say so.
- **Conversion asymmetry, flagged deliberately.** The source-level `excludeFilters` list is preserved across a v1alpha1 round-trip; `pullRequestAuthor` on an *accepting* filter is dropped, exactly like the existing v1alpha2-only filter criteria `conclusion` and `checkName`. Preserving a per-filter criterion would need index-keyed annotation data that breaks if a v1alpha1 client reorders `filters`. A test pins both halves and `docs/reference.md` states it. Happy to add it if you want the guarantee to be uniform.
- **Matching is exact and case-sensitive**, consistent with `excludeAuthors`. GitHub logins are case-insensitive, so `Dependabot[bot]` will not match `dependabot[bot]`; documented rather than diverging from the sibling field.
- v1alpha1's Go types and schema are untouched. All CRD YAML is `make update` output.
- Tests: the positive criterion (match, non-match, and an event with no author); the regression case asserting the same fixture **does** match with only `excludeAuthors` and does **not** once a rule is added; AND-within-a-rule; OR-across-rules; a rule spanning eight event types including `pull_request_review_thread` and the `issues`/`push` non-matches; a scoped rule not leaking to another event type; sender-vs-author independence in both directions; the changed-files fetch being skipped for an excluded event; conversion round-trip plus no-annotation, malformed, and five out-of-schema annotation cases. CEL cannot be unit tested, so there is also an envtest spec that creates the spawner against a real API server and asserts the valid rule stores and round-trips while each invalid one is rejected with the expected message.
- Verified locally: `make verify`, `make test`, and `make test-integration` (152 + 12 specs) pass.

```release-note
Added `pullRequestAuthor` matching and a `spec.when.githubWebhook.excludeFilters` list to TaskSpawner and SessionSpawner. Unlike `excludeAuthors`, which matches the event sender, `pullRequestAuthor` matches the user who opened the pull request an event is about, so a spawner can ignore activity on a bot's own pull requests even when a human triggers the event. Exclusion rules AND their criteria, reject an event whichever accepting filter matched, and may omit `event` to apply across every subscribed event type; events with no pull request author, such as `push` and `issues`, are unaffected. `spec.when.githubWebhook.filters` is now limited to 50 entries.
```


knechtionscoding:
Amended to `49b36f6` (single commit, as this repo squashes for upstream).

**Conversion P2 — valid, fixed.** The restore path decoded the preservation annotation and assigned it without re-checking the field's own constraints. The annotation is user-writable on a v1alpha1 object and the API server does not re-validate the output of a conversion webhook, so a hand-written value could restore a list violating `MaxItems=50` / items `MinLength=1` / `MaxLength=64` straight into v1alpha2. `restorePreservedGitHubWebhookExcludePullRequestAuthors` now validates the decoded list first and ignores it wholesale rather than applying part of it, consistent with how it already treats malformed JSON. Item lengths are counted in runes to match how the API server evaluates OpenAPI `maxLength`, and the bounds are named constants next to the annotation with a comment tying them to the kubebuilder markers.

Tests cover over-`MaxItems`, over-`MaxLength`, an empty item against `MinLength`, and a non-array payload, plus a boundary case (exactly 50 items, longest entry exactly 64 chars) asserting valid data still restores. I verified the tests fail without the guard, so they are not vacuous.

**Webhook P1 — not valid, no change made.** The finding states that `github.ParseWebHook` rejects `pull_request_target` before the raw-JSON fallback runs, so such a spawner cannot match at all. go-github v66 does register the alias — `messages.go:96` maps `"pull_request_target"` to `&PullRequestTargetEvent{}` — so parsing succeeds and the event reaches the `default` arm, which is exactly where the fallback lives. Probed directly:

```
ParseWebHook err=<nil> type=*github.PullRequestTargetEvent
PullRequestAuthor="bot-user" Sender="human" Action="opened"
```

There is also no event-type gate upstream that could reject it: `Events` is a free-form `[]string` with no enum, and the handler passes the raw `X-GitHub-Event` header straight to `ParseGitHubWebhook`. `TestMatchesGitHubEvent_ExcludePullRequestAuthorsAcrossEventTypes/pull_request_target_on_a_bot-authored_PR_is_excluded` covers this end to end and passes. Normalizing the alias to the `pull_request` parser would be a no-op at best. Happy to revisit if you are seeing a failure I have not reproduced.

`make verify`, `make test`, and `make test-integration` (12/12) pass.

gjkim42 (commented):
Matching the PR author independently of the event sender addresses a real gap, and the author extraction in this PR remains useful. My concern is the API shape shared with #1703: adding `excludePullRequestAuthors`, `excludeChannels`, and further `excludeX` fields can give every matching attribute a separately maintained negative counterpart.

I suggest introducing ordinary `pullRequestAuthor` matching criteria and applying exclusion through source-level `excludeFilters`. The positive and negative paths should reuse the same typed criteria and matcher. `GenericWebhook` already uses this pattern with `filters` and `excludeFilters` sharing `GenericWebhookFilter`.

For example, this proposed configuration would cover both PR events and comments on PRs:

```yaml
githubWebhook:
  events: ["pull_request", "issue_comment"]
  excludeFilters:
    - pullRequestAuthor: "dependabot[bot]"
```

Conditions within a rule should be ANDed, and matching any exclusion rule should reject the event regardless of which positive filter matches. Keep exclusions outside the existing OR-based accepting-filter list. A missing PR author should not match, retaining the documented limitation for events such as `check_run` that lack that information.

Existing positive GitHub filters require `event`. Share their underlying matching criteria while allowing exclusion rules to omit the event scope and apply across the subscribed events; preserve the existing positive-filter contract. Invalid rules should surface configuration errors.

Please coordinate the shared shape with #1703, keep shipped fields and their semantics functional, and preserve the new v1alpha2 capability across conversion. This can remain scoped to PR-author matching here; it does not require migrating every existing filter in the same PR.

PR #1726, opened 2026-09-03
feat: support optional skills packages

scottt732:
#### What type of PR is this?

/kind api

Adds an opt-in `spec.skills[].optional` field to the v1alpha2 AgentConfig API.

Required packages keep the existing fail-fast behavior. An optional package logs its source and allows the remaining packages and agent startup to continue when `npx skills add` fails. This prevents an unavailable third-party skills package from blocking an otherwise usable agent when the caller has explicitly accepted degraded capability.

The field defaults to false, is preserved across v1alpha1 conversion through an annotation, and does not add new surface area to the compatibility API version.

N/A

- `make verify` passes.
- `make test` passes.
- `make build` passes after removing a Homebrew `LDFLAGS` value from the local environment; the Makefile forwards that variable to Go's `-ldflags`.
- Script-execution tests cover optional failure with a later successful package, required fail-fast behavior, and the all-optional/all-failed case.
- Conversion tests cover round-trip preservation and ambiguous duplicate protection.

```release-note
AgentConfig v1alpha2 skills can set `optional: true` so a package installation failure is logged without blocking agent startup.
```

---
## Summary by cubic
Adds `spec.skills[].optional` to the v1alpha2 AgentConfig API so agent startup continues when an optional skills package fails to install, while required packages keep the existing fail-fast behavior.

- Optional package failures are logged with the package source; remaining packages and agent startup proceed, and an all-optional failure leaves an empty plugin skills directory.
- Required packages keep failing fast on install, relocation, or install-nothing failures; `secretRef` validation is unchanged and still fails before Job creation. The field defaults to `false` and is preserved across v1alpha1 conversion via an annotation.
- Optional is v1alpha2-only and rejected by the CLI when only v1alpha1 CRDs exist; script-execution and e2e tests cover the continuation and fail-fast paths.

<sup>Written for commit b94c89d3ed2ee3a41bad7e58bbb71d3754d4d232. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/kelos-dev/kelos/pull/1726?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gjkim42 (approved):
Lgtm

@scottt732 
Can you squash commits for this?

scottt732:
@gjkim42 squashed into a single commit and force-pushed — now at `b94c89d`.

The tree is byte-identical to the reviewed head `7336261` (`git diff 7336261 b94c89d` is empty), so this is purely a history cleanup. Your approval carried over to the new head.

